PR #15, opened 2016-09-14
ENH/REF: Upgrade plugin to use new transformers

thermokarst:
- Use new-style views and transformers
- Refactor tests to use new plugin test harness
- Drop QZAs from test data, replace with source data
- Since we weren't using all of the constituent elements of `ReferenceFeatures[SSU]`, I opted for splitting this up into `FeatureData[Taxonomy]` and `FeatureData[*Sequence]` (where applicable).

Note, this plugin still requires the changes proposed in https://github.com/qiime2/qiime2/pull/125. @BenKaehler, your fork is a bit outdated so if you test this locally please make sure you grab the latest changes to qiime2!


thermokarst:
Also should mention that I broke up the SE vs PE methods, instead of patching the function during registration. The split methods are backed by a helper function. This was done in a few places in `_classifier.py`.

PR #47, opened 2017-01-25
ENH: Many fixes and upgrades

BenKaehler:
Fixes qiime2/q2-feature-classifier#44, qiime2/q2-feature-classifier#43, qiime2/q2-feature-classifier#40, qiime2/q2-feature-classifier#39, qiime2/q2-feature-classifier#30, and qiime2/q2-feature-classifier#29. Incorporates other minor enhancements and fixes.

PR #49, opened 2017-01-30
ENH: add version of sklearn in classifier artifact

### LINKED ISSUE #48: Warn users about scikit-learn version mismatch at runtime
For trained DBs, it would be really nice to warn users that they are using a version of scikit-learn that is incompatible with the version used to create the DB. We could do this by updating the DirFmt to encode scikit-learn version, then validate version at runtime.

ebolyen:
An error is now raised requesting that the user reclassify if the version of
sklearn is either not present in the artifact, or if it does not match the
version of sklearn currently installed.

Fixes #48

### DIFF
--- a/q2_feature_classifier/_taxonomic_classifier.py
+++ b/q2_feature_classifier/_taxonomic_classifier.py
@@ -10,6 +10,7 @@ import json
 import tarfile
 import os
 
+import sklearn
 from sklearn.externals import joblib
 import qiime2.plugin
 import qiime2.plugin.model as model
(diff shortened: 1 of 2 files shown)


BenKaehler:
Thanks @ebolyen, sorry I've been unresponsive on this. I had to wait until I had time to have a proper look at it. Thanks very much for making these changes.

I have a few bits of feedback on these changes that I'll summarise here instead of commenting directly on files. I find it difficult to keep track of file comments.

- Why create a `VersionedTaxonomicClassifierDirFmt`? There is no reason to keep the `TaxonomicClassifierDirFmt`, so just change it, I reckon. That also has the advantage that you don't have to change the methods to return the new data type.
- I've made significant changes to `test_taxonomic_classifier.py` in the other open pull request, so some merging will be required. You're certainly a more advanced git user than I am, so I won't offer advice on how to do that.
- I've taken `JSONFormat` out in the other open pull request, because it was unused. You could either put it back in with this one or move it to `q2-types`. It seems like it might be generally useful.

Thanks again.

jairideout:
These changes were ported into #47.

PR #50, opened 2017-02-03
MISC: minor additions based on review of #47

jairideout:
Pair-programmed with @thermokarst and @ebolyen.

### DIFF
--- a/q2_feature_classifier/_cutter.py
+++ b/q2_feature_classifier/_cutter.py
@@ -36,6 +36,12 @@ def _primers_to_regex(f_primer, r_primer):
 def _local_aln(primer, sequence):
     best_score = None
     for one_primer in primer.expand_degenerates():
+        # `sequence` may contain degenerates. These will usually be N
+        # characters, which SSW will score as zero. Although undocumented, SSW
+        # will treat other degenerate characters as a mismatch. We acknowledge
+        # that this approach is a heuristic to finding an optimal alignment and
+        # may be revisited in the future if there's an aligner that explicitly
+        # handles degenerates.
         this_aln = \
             skbio.alignment.local_pairwise_align_ssw(one_primer, sequence)
         score = this_aln[1]
--- a/q2_feature_classifier/_taxonomic_classifier.py
+++ b/q2_feature_classifier/_taxonomic_classifier.py
@@ -95,7 +95,7 @@ def _2(data: Pipeline) -> TaxonomicClassiferTemporaryPickleDirFmt:
 
 
 @plugin.register_transformer
-def _3(dirfmt: TaxonomicClassifierDirFmt) -> dict:
+def _3(dirfmt: TaxonomicClassifierDirFmt) -> Pipeline:
     raise ValueError('The scikit-learn version could not be determined for'
                      ' this artifact, please retrain your classifier for your'
                      ' current deployment to prevent data-corruption errors.')
@@ -117,7 +117,7 @@ def _5(data: dict) -> JSONFormat:
 
 # Registrations
 plugin.register_semantic_types(TaxonomicClassifier)
-plugin.register_formats(PickleFormat, TaxonomicClassifierDirFmt,
+plugin.register_formats(TaxonomicClassifierDirFmt,
                         TaxonomicClassiferTemporaryPickleDirFmt)
 plugin.register_semantic_type_to_format(
     TaxonomicClassifier,
--- a/q2_feature_classifier/tests/test_taxonomic_classifier.py
+++ b/q2_feature_classifier/tests/test_taxonomic_classifier.py
@@ -80,7 +80,7 @@ class TestTransformers(TaxonomicClassifierTestBase):
             transformer(input)
 
     def test_old_dirfmt(self):
-        transformer = self.get_transformer(TaxonomicClassifierDirFmt, dict)
+        transformer = self.get_transformer(TaxonomicClassifierDirFmt, Pipeline)
         with open(os.path.join(self.temp_dir.name,
                                'preprocess_params.json'), 'w') as fh:
             fh.write(json.dumps([]))


jairideout:
@BenKaehler, if you're around to review this soonish, we'd appreciate it! If not, we'll have @gregcaporaso review later today. I think it's Saturday morning your time, so no worries if you can't get to it!

q2_feature_classifier/_cutter.py
```diff
@@ -36,6 +36,8 @@ def _primers_to_regex(f_primer, r_primer):
 def _local_aln(primer, sequence):
     best_score = None
     for one_primer in primer.expand_degenerates():
+        # `sequence` may contain degenerates. Although undocumented, SSW will
+        # treat each degenerate character as a mismatch.
```
BenKaehler: Should read:
```
# `primer` may contain degenerates. Although undocumented, SSW will
# treat each degenerate character as a mismatch
```

ebolyen: Actually, this comment is specifically about `sequence`, we're still expanding the degenerates in the primer, but we're commenting to purposefully say if there are degenerates in the reference, that we don't care and a mismatch should still be "good enough".

q2_feature_classifier/_taxonomic_classifier.py
```diff
@@ -95,7 +95,7 @@ def _2(data: Pipeline) -> TaxonomicClassiferTemporaryPickleDirFmt:
 
 
 @plugin.register_transformer
-def _3(dirfmt: TaxonomicClassifierDirFmt) -> dict:
+def _3(dirfmt: TaxonomicClassifierDirFmt) -> Pipeline:
```
BenKaehler: I'll take your authority on this one, because my understanding of the transformer architecture is superficial. However, the transformer used to return a `dict`, and this will never return a `Pipeline` because it will raise a `ValueError`. As I said however, you're the experts.

q2_feature_classifier/_taxonomic_classifier.py
```diff
@@ -117,7 +117,7 @@ def _5(data: dict) -> JSONFormat:
 
 # Registrations
 plugin.register_semantic_types(TaxonomicClassifier)
-plugin.register_formats(PickleFormat, TaxonomicClassifierDirFmt,
```
BenKaehler: Good call.

q2_feature_classifier/_taxonomic_classifier.py
ebolyen: QIIME 2 has to search for the transformer before it calls it, and `Pipeline` is part of the key it uses to identify this. Since the method wants a `Pipeline` it needs to find a transformer turning the artifact's view into that.

q2_feature_classifier/custom.py
```diff
@@ -15,7 +15,11 @@
 
 class LowMemoryMultinomialNB(MultinomialNB):
     def __init__(self, alpha=1.0, fit_prior=True, class_prior=None,
-                 chunk_size=-1):
+                 chunk_size=0):
```
BenKaehler: This introduces an inconsistency with, for instance, the `length` parameter for `extract_reads`. I like the negative values because they're clearly not a valid `chunk_size` or read `length`.

q2_feature_classifier/tests/test_taxonomic_classifier.py
```diff
@@ -80,7 +80,7 @@ def test_old_sklearn_version(self):
             transformer(input)
 
     def test_old_dirfmt(self):
-        transformer = self.get_transformer(TaxonomicClassifierDirFmt, dict)
+        transformer = self.get_transformer(TaxonomicClassifierDirFmt, Pipeline)
```
BenKaehler: This seems funny to me, but happy to except your authority.

q2_feature_classifier/_cutter.py
BenKaehler: Ok, in that case, I'm not sure that that's quite correct. The only degenerates in `sequence` should be `N` characters, and from memory SSW treats them specially, allocating them a score of zero.

q2_feature_classifier/_taxonomic_classifier.py
jairideout: Since the method now expects a `Pipeline` instead of a `dict`, we need a transformer from the old (deprecated) format into `Pipeline`. That way if a user tries running the method with a .qza in the old format, they'll get a `ValueError`. If we left the return type as `dict`, that transformer wouldn't get invoked and the user would instead get an error stating that the old format doesn't have a transformer into `Pipeline`. With this change to `Pipeline`, users will get a more informative error message about the deprecated format instead of a missing transformer.

BenKaehler: Thanks for the explanation!

q2_feature_classifier/tests/test_taxonomic_classifier.py
jairideout: See my comment about why this change is necessary: https://github.com/qiime2/q2-feature-classifier/pull/50#discussion_r99434658

q2_feature_classifier/custom.py
BenKaehler: As Jai has commented elsewhere, I think the solution here for the moment is better documentation.

jairideout: OK we'll revert to keep this parameter consistent with the others. We were trying to avoid having two sentinels (-1 and 0, which both do the same thing) because it's likely to raise questions from users and we really only need one sentinel. But it's probably worth tackling this later so all parameters get the same updates.

For now, it'd be helpful to document this behavior in the parameter help text whenever that gets added. Thanks!

q2_feature_classifier/tests/test_taxonomic_classifier.py
BenKaehler: Thanks Jai, no problems.

q2_feature_classifier/_cutter.py
BenKaehler: From [_ssw_wrapper.pyx](https://github.com/biocore/scikit-bio/blob/master/skbio/alignment/_ssw_wrapper.pyx):
```
    cdef cnp.ndarray[cnp.int8_t, ndim = 1, mode = "c"] \
            _build_match_matrix(self, match_score, mismatch_score):
        sequence_order = "ACGTN"
        dict2d = {}
        for row in sequence_order:
            dict2d[row] = {}
            for column in sequence_order:
                if column == 'N' or row == 'N':
                    dict2d[row][column] = 0
                else:
                    dict2d[row][column] = match_score if row == column \
                        else mismatch_score
        return self._convert_dict2d_to_matrix(dict2d)
```

jairideout: Thanks @BenKaehler! I'll get that comment updated.

q2_feature_classifier/_cutter.py
```diff
@@ -36,6 +36,12 @@ def _primers_to_regex(f_primer, r_primer):
 def _local_aln(primer, sequence):
     best_score = None
     for one_primer in primer.expand_degenerates():
+        # `sequence` may contain degenerates. These will usually be N
+        # characters, which SSW will score as zero. Although undocumented, SSW
+        # will treat other degenerate characters as a mismatch. We acknowledge
+        # that this approach is a heuristic to finding an optimal alignment and
+        # may be revisited in the future if there's an aligner that explicitly
+        # handles degenerates.
```
BenKaehler: 👍

jairideout:
Thanks for the review @BenKaehler! We addressed your comments. Please let us know if there's anything else, otherwise feel free to merge.

jairideout:
Thanks @BenKaehler!

PR #55, opened 2017-02-09
vsearch and blast plugin methods and tests

nbokulich:


.travis.yml
```diff
@@ -15,6 +15,8 @@ install:
   # dependencies via pip.
   - conda create --yes -n test-env -c biocore python=$PYTHON_VERSION scikit-bio jupyter click nose scikit-learn
   - source activate test-env
+  - conda install --yes -c bioconda blast=2.6.0
```
gregcaporaso: Is this BLAST+ or legacy BLAST? We probably should use BLAST+ in QIIME 2, since legacy BLAST is pretty outdated at this point.

q2_feature_classifier/_blast.py
```diff
@@ -0,0 +1,77 @@
...
+                                    _run_command)
+
+
+def blast(query: DNAFASTAFormat, reference_reads: DNAFASTAFormat,
+          reference_taxonomy: pd.Series, maxaccepts: int=10, min_id: int=0.8,
+          strand: str='both', evalue: float=0.001, t_delim: str=';',
+          min_consensus: float=0.51, unassignable_label: str="Unassigned",
+          num_threads: str=1, params: str=None) -> pd.DataFrame:
+
+    min_id = min_id * 100
```
gregcaporaso: You should raise an error here if `min_id` is outside the range of [0.0, 1.0]. A common mistake is that users will pass a percentage (e.g., 97) rather than a fractional value (e.g., 0.97). While this check isn't perfect, it'd catch most instances. You should also include examples of how to use it in the help text when you register the method, to help users avoid this mistake.

q2_feature_classifier/_blast.py
```diff
@@ -0,0 +1,77 @@
...
+            'reference_reads': FeatureData[Sequence],
+            'reference_taxonomy': FeatureData[Taxonomy]},
+    parameters={'evalue': Float, 'maxaccepts': Int, 'min_id': Float,
+                'strand': Str, 't_delim': Str, 'params': Str,
+                'min_consensus': Float, 'unassignable_label': Str,
+                'num_threads': Int},
+    outputs=[('classification', FeatureData[Taxonomy])],
+    name='Assign taxonomy using BLAST+.',
+    description='Assign taxonomy to query sequences using BLAST+.',
+    parameter_descriptions={'evalue': 'BLAST expectation value (E) threshold'
```
gregcaporaso: Can you format this as illustrated in the example [here](https://docs.qiime2.org/2017.2/plugins/developing/#registering-a-method). It's not essential, but it makes this a lot more readable. 

Also, can you add `input_descriptions` and `output_descriptions`?

q2_feature_classifier/_blast.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def blast(query: DNAFASTAFormat, reference_reads: DNAFASTAFormat,
+          reference_taxonomy: pd.Series, maxaccepts: int=10, min_id: int=0.8,
+          strand: str='both', evalue: float=0.001, t_delim: str=';',
+          min_consensus: float=0.51, unassignable_label: str="Unassigned",
+          num_threads: str=1, params: str=None) -> pd.DataFrame:
+
+    min_id = min_id * 100
+    params = _parse_params(params)
```
gregcaporaso: This makes me a bit nervous. Do you think it's essential? It might represent a security issue (for example, what would happen if a user provided `;,rm,-rf,/`? (Don't try though, because I have no idea what would happen!) It also can be confusing - for example, does providing `evalue` here override the `evalue` parameter?

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,229 @@
...
+            The label to apply if no acceptable annotations are identified.
+        Result
+        ------
+        consensus_annotation
+            List containing the consensus assignment
+        consensus_fraction
+            Fraction of input annotations that agreed at the deepest
+            level of assignment
+    """
+    # This code has been ported from QIIME 1.9.1 with
```
gregcaporaso: Can you bump this up to just above the function definition (line 142 of this file)?

q2_feature_classifier/_blast.py
```diff
@@ -0,0 +1,77 @@
...
+from qiime2.plugin import Int, Str, Float
+from .plugin_setup import plugin
+from ._consensus_assignment import (_consensus_assignments,
+                                    _parse_params,
+                                    _run_command)
+
+
+def blast(query: DNAFASTAFormat, reference_reads: DNAFASTAFormat,
+          reference_taxonomy: pd.Series, maxaccepts: int=10, min_id: int=0.8,
+          strand: str='both', evalue: float=0.001, t_delim: str=';',
```
gregcaporaso: `t_delim` shouldn't need to be in here, as the format used in qiime2 specifies that these are semi-colon-separated.

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,229 @@
...
+import tempfile
+import subprocess
+import pandas as pd
+from os.path import isfile
+from collections import Counter
+
+
+def _consensus_assignments(cmd, ref_taxa, t_delim=';', min_consensus=0.51,
+                           unassignable_label="Unassigned"):
+    '''Run command on CL and find consensus taxonomy.'''
```
gregcaporaso: Can you expand `CL` rather than abbreviate it (I'm not sure what you're referring to here).

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,229 @@
...
+        query_annotations : dict of lists
+            Keys are query identifiers, and values are lists of all
+            taxonomic annotations associated with that identfier.
+        Returns
+        -------
+        dict
+            Keys are query identifiers, and values are the consensus of the
+            input taxonomic annotations.
+    """
+    # This code has been ported and adapted from QIIME 1.9.1 with
```
gregcaporaso: Can you bump this up to just above the function definition (line 114)?

q2_feature_classifier/_vsearch.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+plugin.methods.register_function(
+    function=vsearch,
+    inputs={'query': FeatureData[Sequence],
+            'reference_reads': FeatureData[Sequence],
+            'reference_taxonomy': FeatureData[Taxonomy]},
+    parameters={'maxaccepts': Int, 'min_id': Float, 'strand': Str,
+                't_delim': Str, 'params': Str, 'min_consensus': Float,
+                'unassignable_label': Str, 'num_threads': Int},
```
gregcaporaso: Should the default for `unassignable_label` be the same for all classifiers in this package? You could define a function that returns this, for example:

```python
def _get_default_unassignable_label():
    return "Unassigned"
```

.travis.yml
nbokulich: Thanks for asking! Yes, this is ``blast+``, though the package name is not clear. See the package info [here](https://anaconda.org/bioconda/blast). version 2.6.0 is the latest version, released in late January 2017.

q2_feature_classifier/_blast.py
nbokulich: non-essential, removing.

q2_feature_classifier/_blast.py
nbokulich: ``_validate_params``, at line 47 of ``_consensus_assignment.py``, now tests this and other parameters.

q2_feature_classifier/_blast.py
nbokulich: Re-formatted at lines 53+, added ``input_descriptions`` and ``output_descriptions`` at lines 62 and 78.

q2_feature_classifier/_consensus_assignment.py
nbokulich: line 158

q2_feature_classifier/_blast.py
nbokulich: removed ``f_delim`` ('\t') and ``t_delim`` (';'), now hardcoded.

q2_feature_classifier/_consensus_assignment.py
nbokulich: line 24

q2_feature_classifier/_consensus_assignment.py
nbokulich: line 131

q2_feature_classifier/_vsearch.py
nbokulich: ``_get_default_unassignable_label `` at line 16

this is now called to set default in all applicable functions.

nbokulich:
Thanks @gregcaporaso ! I have made all requested changes.

Coverage is > 95% for all modules included in this PR (output of ``nosetests --with-coverage`` below):

```
Name                                                       Stmts   Miss  Cover
------------------------------------------------------------------------------
q2_feature_classifier/_blast.py                               20      0   100%
q2_feature_classifier/_consensus_assignment.py                92      3    97%
q2_feature_classifier/_vsearch.py                             13      0   100%
q2_feature_classifier/tests/test_consensus_assignment.py     123      0   100%
```


A couple other small changes in there cover minor bugs I uncovered (``blast`` does not output any results if there are no hits for any query seqs, throwing an error when attempting to fit any empty dict into a 2-column ``pandas.Series``)

gregcaporaso:
That's great, thanks @nbokulich! I'll review this ASAP.

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,243 @@
...
+    ref_taxa: dict or pd.Series
+        Reference taxonomies in tab-delimited format:
+            <accession ID>  kingdom;phylum;class;order;family;genus;species
+    '''
+    obs_taxa = {}
+    lines = _open_list_or_file(assignments)
+
+    for line in lines:
+        if not line.startswith('#') or line == "":
+            i = line.split('\t')
```
gregcaporaso: Doesn't really matter all that much, but this would be a better way to do this:

```
fields = line.split('\t') 
```

`i` is traditionally used to represent an index, so it's just a little confusing to use it this way.

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,243 @@
...
+            # ref taxonomy IDs get imported as a str when using CLI
+            # but imported as int in python interpreter (e.g., during tests)
+            # the following allows dict lookup as str or int.
+            try:
```
gregcaporaso: Using `if`/`else` statements instead of `try`/`except` is typically more clear. You also typically want to use `try`/`expect` for "unusual" cases, and `if`/`else` for cases that frequently come up. You could do this as:

```python
if fields[1] == '*':
    t = [unassignable_label]
else:
    try:
        t = ref_taxa[fields[1]].split(';')
    except KeyError:
        raise KeyError('Identifier %s was reported in taxonomic search results, but was not present in the reference taxonomy.' % fields[1])
```

For this to work, you'll need to ensure that the indices in `ref_taxa` are always strings. Is that possible?

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,243 @@
...
+                t = ref_taxa[i[1]].split(';')
+            except (KeyError, TypeError):
+                try:
+                    t = ref_taxa[int(i[1])].split(';')
+                # if vsearch fails to find assignment, it reports '*' as the
```
gregcaporaso: This case looks like it's vsearch related, but this function (at least based on its name) is blast-related.

gregcaporaso: Oh - is that because this is the BLAST format generated by vsearch?

nbokulich: Yep, vsearch is generating BLAST-format results.

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,243 @@
...
+    exp = [line.strip('>') for line in exp if line.startswith('>')]
+
+    # iterate over expected IDs, add as unassigned to obs_taxa if missing
+    for _id in exp:
+        if _id not in obs_taxa:
+            obs_taxa[_id] = (unassignable_label, 0.0)
+
+    return obs_taxa
+
+
```
gregcaporaso: Can you add the following note here:

```
# Replace this function with QIIME2 API for wrapping commands/binaries,
# pending https://github.com/qiime2/qiime2/issues/224
```

q2_feature_classifier/_vsearch.py
```diff
@@ -0,0 +1,64 @@
...
+        unassignable_label=unassignable_label)
+    return consensus
+
+
+plugin.methods.register_function(
+    function=vsearch,
+    inputs={'query': FeatureData[Sequence],
+            'reference_reads': FeatureData[Sequence],
+            'reference_taxonomy': FeatureData[Taxonomy]},
+    parameters={'maxaccepts': Int, 'min_id': Float, 'strand': Str,
```
gregcaporaso: Can you make `strand` into `Str % Choices(['both', ...])`, where `...` is the other choices?

q2_feature_classifier/_consensus_assignment.py
nbokulich: The indices (accession ids) in ``ref_taxa`` are frequently numeric, and get automatically cast as integers. This is all done internally by ``q2-types`` (I believe) when ``ref_taxa`` is imported as a ``pandas.Series``, so I cannot control that behavior here. The best solution I could come up with previously was to convert to string — and I could still evaluate this in an if/else statement if that's acceptable.

nbokulich: Never mind, something like ``isinstance(fields[1], str)`` does not help us here, as ``fields`` will always be a string — it is ``ref_taxa[fields[1]]`` that we need to evaluate, but that evaluation will raise an error if the indices in ``ref_taxa`` are not strings. @gregcaporaso is the following acceptable, or is there a better way to evaluate the whether all indices in ``ref_taxa`` are of type string, in which case I can use an if/else instead of the try/except here:

```
if fields[1] == '*':
    t = [unassignable_label]
else:
    try:
        t = ref_taxa[fields[1]].split(';')
    except (KeyError, TypeError):
        t = ref_taxa[int(fields[1])].split(';')
```

q2_feature_classifier/_vsearch.py
nbokulich: Done. changed in ``_blast.py`` and ``_vsearch.py``

q2_feature_classifier/_consensus_assignment.py
nbokulich: Done

q2_feature_classifier/_consensus_assignment.py
nbokulich: Done.

q2_feature_classifier/_vsearch.py
```diff
@@ -0,0 +1,64 @@
...
+# ----------------------------------------------------------------------------
+
+import pandas as pd
+from ._consensus_assignment import (_consensus_assignments,
+                                    _get_default_unassignable_label,
+                                    _validate_params)
+from q2_types.feature_data import (
+    FeatureData, Taxonomy, Sequence, DNAFASTAFormat)
+from .plugin_setup import plugin
+from qiime2.plugin import Int, Str, Float
```
gregcaporaso: add `Choices, Range` to this last import (see my change below which will use these).

q2_feature_classifier/_vsearch.py
gregcaporaso: Hold off on adding the `Range` parameters for a minute. This may require a change to q2cli.

q2_feature_classifier/_vsearch.py
```diff
@@ -0,0 +1,64 @@
...
+            'reference_taxonomy': FeatureData[Taxonomy]},
+    parameters={'maxaccepts': Int, 'min_id': Float, 'strand': Str,
+                'min_consensus': Float, 'unassignable_label': Str,
+                'num_threads': Int},
+    outputs=[('classification', FeatureData[Taxonomy])],
+    input_descriptions={'query': 'Sequences to classify taxonomically.',
+                        'reference_reads': 'reference sequences.',
+                        'reference_taxonomy': 'reference taxonomy labels.'},
+    parameter_descriptions={
+        'strand': 'plus|both',
```
gregcaporaso: Can you make this description more descriptive (say what it does). When you update the type of this parameter to include `Choices`, you won't need to describe the choices in the help text anymore as those will be included by default.

q2_feature_classifier/_vsearch.py
gregcaporaso: You should make these changes (and similar changes in the `blast` method, as necessary), but note that you'll now depend on master of `qiime2` and `q2cli` (those required some changes to support these range parameters).

q2_feature_classifier/_vsearch.py
gregcaporaso: Note however that you will need to include the range boundaries in the descriptions for the range parameters. Those do not currently get displayed by click (the CLI framework that we're using).

q2_feature_classifier/_vsearch.py
```diff
@@ -0,0 +1,64 @@
...
+        'maxaccepts': ('Maximum number of hits to keep for each query. Must be'
+                       '>= 1'),
```
gregcaporaso: Missing a space between *be* and *>=*.

gregcaporaso: For consistency with the other ranges that you're describing here, could you make this say: 

*Must be in range [0, infinity).*

q2_feature_classifier/_vsearch.py
```diff
@@ -0,0 +1,64 @@
...
+        'min_id': 'Reject match if percent identity to query is lower. Range '
```
gregcaporaso: *Range* -> *Must be in range*

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,243 @@
...
+        # to empty result file (must create empty dict or fail on pd error)
+        if not consensus:
+            consensus = {'': ('', '')}
+        result = pd.DataFrame.from_dict(consensus, 'index')
+        result.index.name = 'Feature ID'
+        result.columns = ['Taxon', 'Confidence']
+        return result
+
+
+def _validate_params(min_id, maxaccepts, min_consensus):
```
gregcaporaso: You can drop this function now, as well as all related tests. The `Range` parameters that I'm having you add will take care of this validation for you.

q2_feature_classifier/_blast.py
```diff
@@ -0,0 +1,80 @@
...
+
+    return consensus
+
+
+plugin.methods.register_function(
+    function=blast,
+    inputs={'query': FeatureData[Sequence],
+            'reference_reads': FeatureData[Sequence],
+            'reference_taxonomy': FeatureData[Taxonomy]},
+    parameters={'evalue': Float, 'maxaccepts': Int, 'min_id': Float,
```
gregcaporaso: See the `Choice` and `Range` parameters that I'm having you add to the `vsearch` function - you should use those here as well.

q2_feature_classifier/_consensus_assignment.py
gregcaporaso: @nbokulich, makes sense. We're going to update these indices in the future to always be strings, but you're right that you can't rely on that right now. 

Will `ref_taxa` always be a series? If so, you should be able to do:

```
if ref_taxa.index.dtype == 'int':
   id_ = int(fields[1])
else: 
    id_ = fields[1]

try:
    t = ref_taxa[id_]
except KeyError:
    raise KeyError('Identifier %s was reported in taxonomic search results, but was not present in the reference taxonomy.' % str(id_))

try: 
    t = t.split(';')
except ValueError:
    # I think this is the case that you're catching a ValueError for
    raise ValueError('Reference taxonomy %s (id: %s) is incorrectly formatted.' % (t, str(id_)))
```

This behaves very explicitly, and will allow you to raise the most relevant errors.

nbokulich:
Thanks @gregcaporaso ! I have added all additional changes you requested, and made the same changes to ``_blast.py`` that you recommended for ``_vsearch.py``. It is ready if you wanted to take another look, otherwise it is all ready for @BenKaehler to review.

gregcaporaso:
@nbokulich, you're missing a space between *in* and *range* and *hit* and *to* in the vsearch do for *min-consensus*. Could you add that?

```
Minimum fraction of
assignments must match top hitto be accepted
as consensus assignment. Must be inrange [0.51
- 1.0].
```

This also happens in your description of the BLAST+ method:

```
PerformsBLAST+
```

Note that when you join strings in the way that we do here, there is no space inserted between the pieces of the strings, so you have to add that yourself. By convention, I always put the space at the end of the line, for example:

```python
("This is an "
 "example "
 "of a multi-line "
 "string.")
```

Can you review all of your strings and make sure they are formatted to include spaces? A good way to spot check this is just to read all of the docs when calling the new methods with ``--help`` on the command line - the missing spaces jump out at you then. 

Also, I wouldn't request this on it's own, but since you're making this change, could you update all of the ranges to be specified using interval notation as described [here](https://en.wikipedia.org/wiki/Interval_(mathematics)). 

For example, *min-consensus* would be `[0.5, 1.0]`.

nbokulich:
Thanks @gregcaporaso ! I have changed to interval notation and fixed the missing spaces.

gregcaporaso:
Looks good, thanks @nbokulich! I've also installed and run both of the new classifiers on Linux and macOS, so I can confirm that the results look reasonable and that the conda packages work in both environments. 

We'll wait for @BenKaehler's review, and then this is ready for merge.

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,252 @@
...
+                        'formatted.').format(t, str(id_)))
+
+            if fields[0] in obs_taxa.keys():
+                obs_taxa[fields[0]].append(t)
+            else:
+                obs_taxa[fields[0]] = [t]
+    return obs_taxa
+
+
+def _open_list_or_file(infile):
```
BenKaehler: Is `_open_list_or_file` ever passed a file?

q2_feature_classifier/_vsearch.py
```diff
@@ -0,0 +1,71 @@
...
+            strand: str='both', min_consensus: float=0.51,
+            unassignable_label: str=_get_default_unassignable_label(),
+            num_threads: str=1) -> pd.DataFrame:
+
+    seqs_fp = str(query)
+    ref_fp = str(reference_reads)
+    cmd = ['vsearch', '--usearch_global', seqs_fp, '--id', str(min_id),
+           '--strand', strand, '--maxaccepts', str(maxaccepts),
+           '--maxrejects', '0', '--output_no_hits', '--db', ref_fp,
+           '--threads', str(num_threads), '--blast6out']
```
BenKaehler: Again, why rename `id` and `threads`?

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,252 @@
...
+              "messages to stdout and/or stderr.")
+        print("The command being run is below. This command cannot "
+              "be manually re-run as it will depend on temporary files that "
+              "no longer exist.")
+        print("\nCommand:", end=' ')
+        print(" ".join(cmd), end='\n\n')
+    subprocess.run(cmd, check=True)
+
+
+def _import_blast_assignments(
```
gregcaporaso: I was also confused about this. @nbokulich, maybe a better name would be ``_assignments_from_bl7``, and then add a comment inside the top of the function noting that `bl7` refers to the BLAST+ output format 7?

q2_feature_classifier/_blast.py
```diff
@@ -0,0 +1,88 @@
...
+    ref_fp = str(reference_reads)
+    with tempfile.TemporaryDirectory() as blastdir:
+        blastref = join(blastdir, 'blastref')
+        # make temporary blast db
+        cmd = ['makeblastdb', '-in', ref_fp, '-out', blastref, '-dbtype',
+               'nucl']
+        _run_command(cmd)
+        # execute blastn search
+        cmd = ['blastn', '-query', seqs_fp, '-evalue', str(evalue), '-strand',
+               strand, '-outfmt', '7', '-db', blastref, '-perc_identity',
```
nbokulich: Was trying to standardize naming between vsearch and blast at some point but that is no longer necessary.

Have changed to ``perc_identity``

q2_feature_classifier/_consensus_assignment.py
nbokulich: I have changed this to ``_import_blast_format_assignments`` to clarify. vsearch is outputting blast6 format. blast6 or blast7 (same thing with/without comment lines) should both be supported so a more generic name than ``_assignments_from_bl7`` seems appropriate, but following the same reasoning I could be more specific here, e.g., ``_assignments_from_bl6_or_bl7``

q2_feature_classifier/_consensus_assignment.py
nbokulich: Yes. The assignments are first output to a temp file by the blast/vsearch subprocess. This function is used to import this to a list — but is also convenient for unit testing, where I can test this function without importing a data file, hence my desire for flexibility here.

q2_feature_classifier/_vsearch.py
nbokulich: ``num_threads`` was an attempt at standardizing nomenclature with ``blast``, as described above, now irrelevant and renamed ``threads``.

``id`` appears to be a built-in function in python so I used ``min_id`` instead of being a naughty programmer. I have renamed this ``id_``, which I've seen others use elsewhere. Is that okay?

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -0,0 +1,252 @@
...
+                        'taxonomy.').format(str(id_)))
+
+                try:
+                    t = t.split(';')
+                except ValueError:
+                    raise ValueError((
+                        'Reference taxonomy {0} (id: {1}) is incorrectly '
+                        'formatted.').format(t, str(id_)))
+
+            if fields[0] in obs_taxa.keys():
```
nbokulich: thanks! that does remove a few lines...

q2_feature_classifier/_blast.py
```diff
@@ -0,0 +1,88 @@
...
+          unassignable_label: str=_get_default_unassignable_label(),
+          num_threads: str=1) -> pd.DataFrame:
+
+    min_id = min_id * 100
+    seqs_fp = str(query)
+    ref_fp = str(reference_reads)
+    with tempfile.TemporaryDirectory() as blastdir:
+        blastref = join(blastdir, 'blastref')
+        # make temporary blast db
+        cmd = ['makeblastdb', '-in', ref_fp, '-out', blastref, '-dbtype',
```
nbokulich: Evidently you are the 💥 📄 , because that works @BenKaehler . Thanks! Now I know who to ask when I'm too lazy to peruse the blast documentation more carefully in the future!

nbokulich:
Thanks @BenKaehler for the review! thermokarst would be proud of your attention to detail. and thanks @gregcaporaso for additional comments. I have made all changes, as detailed above, but pls let me know if my new names are still not clear enough.

:blastlegend: should be its own emoji.

gregcaporaso:
Thanks guys, excited to have this in!

Issue #59, opened 2017-03-31
update method names, now that we have more classifiers


gregcaporaso:
On the dev call we discussed names like:

 - `blast-consensus-classify`
 - `vsearch-consensus-classify`
 - `skl-classify`
 - `skl-fit-classifier`

BenKaehler:
So the current set of commands are
```
  blast                       BLAST+ consensus taxonomy classifier
  classify                    Classify reads by taxon.
  extract-reads               Extract reads from reference.
  fit-classifier              Train a scikit-learn classifier.
  fit-classifier-naive-bayes  Train the naive_bayes classifier.
  vsearch                     VSEARCH consensus taxonomy classifier
```
I will follow perhaps @ebolyen's suggestion that we label the commands in the French style so that group nicely:
```
  classify-sklearn		Pre-fitted sklearn-based taxonomy classifier.
  classify-consensus-blast	BLAST+ consensus taxonomy classifier
  classify-consensus-vsearch	VSEARCH consensus taxonomy classifier
  extract-reads			Extract reads from reference.
  fit-classifier-sklearn	Train an almost arbitrary scikit-learn classifier.
  fit-classifier-naive-bayes	Train a scikit-learn naive Bayes classifier.
```
In future, when we abolish the saving-down of arbitrary trained classifiers, the commands will then become:
```
  classify-sklearn		Pre-fitted sklearn-based taxonomy classifier.
  classify-sklearn-custom	Almost arbitrary sklearn-based taxonomy classifier
  classify-consensus-blast	BLAST+ consensus taxonomy classifier
  classify-consensus-vsearch	VSEARCH consensus taxonomy classifier
  extract-reads			Extract reads from reference.
  fit-classifier-naive-bayes	Train a scikit-learn naive Bayes classifier.
```

Issue #58, opened 2017-03-29
naive bayes classifiers don't work if reads are reverse complement relative to the reference database


gregcaporaso:
I ran into some very strange results when analyzing 454 data that I imported at the `FeatureTable[Frequency]` and `FeatureData[Sequence]` stage. These are skin samples from the [Fierer et al "Keyboard Study"](http://www.pnas.org/content/107/14/6477.full). 

My initial taxonomy barplots looked like this, which are clearly not the right composition:

![screenshot 2017-03-29 14 39 16](https://cloud.githubusercontent.com/assets/192372/24477960/e7b746bc-148d-11e7-9dca-65bc0200b9b1.png)

After spending some time figuring out what was going on, and noticing that the `blast+` assigner in the current development version was giving correct results, I realized that in the `FeatureData[Sequence]` data that I was importing, the sequences are actually the reverse complement relative to what's in Greengenes. I reverse complemented all of my sequences before importing, and the resulting taxa compositions look correct:

![screenshot 2017-03-29 14 39 23](https://cloud.githubusercontent.com/assets/192372/24478035/362211c4-148e-11e7-971e-ed25b9c17238.png)

`blast+` is handling this because it searches for both forward/forward and forward/reverse matches, and I think RDP classifier handles it by training on the forward and reverse complement kmers. @nbokulich, @BenKaehler, does that sound right to you? 

Could we update the naive bayes classifiers, and any others that would have this same issue, to work both for reads that are forward and reverse complement relative to the training data?

BenKaehler:
Thanks @gregcaporaso, according to Wang (2007), for the online version of the RDP classifier at least,

> The Classifier automatically checks both the forward and reverse orientations of the sequence and returns only the results for the correct orientation.

I think this is probably the correct approach, that is to attempt to classify the forward and reverse reads, rather than to make the classifier's job even harder by throwing the reverse complements at it as well in the training step.

Is it valid to assume that all of the reads in a sample are forward or all of the reads are reversed? We could check the first 1000 in both directions, then pick the direction that's giving more "confident" classifications for the rest of the run.

nbokulich:
I think that most cases will have all forward or all reverse reads. I can imagine a scenario where mixed 16S orientations exist, but this doesn't fit any current library prep method that I know, so don't think it is necessary to support — just checking the first N reads makes sense as long as this is documented well.

[...]

Issue #57, opened 2017-03-27
`fit-classifier-naive-bayes` default chunk size to 20000


gregcaporaso:
Discussed this with @BenKaehler on Slack - this will speed up the classifier and avoid memory errors.

jairideout:
Woops, tagged the wrong Ben!

cc: @BenKaehler

gregcaporaso:
Thanks @jairideout!

gregcaporaso:
@BenKaehler, I got an error using a chunk size of 20000 this morning:

```
Plugin error from feature-classifier:

  index out of bounds: 0 <= 0 <= 855, 0 <= 20000 <= 855, 0 <= 20000

Re-run with --verbose to see debug info.
```

I was able to correct this easily my self (by just going with the default of -1), but if we make 20000 the default we'll need to handle this case so users don't get the error - basically we'll just need to detect when 20k isn't a sensible value, and go with something that is.

Issue #53, opened 2017-02-08
Tidy up a little bit of logic


BenKaehler:
The logic on line 118 of `_cutter.py` could be tidied up a bit. Some changes that occurred during the last PR allow an `if` statement to be removed.

PR #62, opened 2017-04-08
FIX: a variety of small enhancements and fixes

BenKaehler:
Fixes qiime2/q2-feature-classifier#59
Fixes qiime2/q2-feature-classifier#58
Fixes qiime2/q2-feature-classifier#57
Fixes qiime2/q2-feature-classifier#53

Addresses most of qiime2/q2-feature-classifier#54

### DIFF
--- a/q2_feature_classifier/_skl.py
+++ b/q2_feature_classifier/_skl.py
@@ -21,8 +21,7 @@ _specific_fitters = [
             'non_negative': True}],
           ['classify',
            {'__type__': 'custom.LowMemoryMultinomialNB',
-            'alpha': 0.01,
-            'chunk_size': 20000}]]]]
+            'alpha': 0.01}]]]]
 
 
 def fit_pipeline(reads, taxonomy, pipeline):
--- a/q2_feature_classifier/_skl.py
+++ b/q2_feature_classifier/_skl.py
@@ -21,7 +21,8 @@ _specific_fitters = [
             'non_negative': True}],
           ['classify',
            {'__type__': 'custom.LowMemoryMultinomialNB',
-            'alpha': 0.01}]]]]
+            'alpha': 0.01,
+            'chunk_size': 20000}]]]]
 
 
 def fit_pipeline(reads, taxonomy, pipeline):
@@ -38,8 +39,6 @@ def _extract_reads(reads):
 
 def predict(reads, pipeline, separator=';', chunk_size=262144, n_jobs=1,
             pre_dispatch='2*n_jobs', confidence=-1.):
-    if confidence >= 0.:
-        chunk_size = chunk_size // 101 + 1
     return (m for c in Parallel(n_jobs=n_jobs, batch_size=1,
                                 pre_dispatch=pre_dispatch)
             (delayed(_predict_chunk)(pipeline, separator, confidence, chunk)
@@ -91,6 +90,7 @@ def _predict_chunk_with_conf(pipeline, separator, confidence, chunk):
 
 
 def _chunks(reads, chunk_size):
+    reads = iter(reads)
     while True:
         chunk = list(islice(reads, chunk_size))
         if len(chunk) == 0:
--- a/q2_feature_classifier/classifier.py
+++ b/q2_feature_classifier/classifier.py
@@ -17,6 +17,7 @@ from qiime2.plugin import Int, Str, Float, Bool, Choices
 from q2_types.feature_data import FeatureData, Taxonomy, Sequence, DNAIterator
 from sklearn.pipeline import Pipeline
 import sklearn
+from numpy import median, array
 
 from ._skl import fit_pipeline, predict, _specific_fitters
 from ._taxonomic_classifier import TaxonomicClassifier
@@ -116,20 +117,20 @@ plugin.methods.register_function(
 )
 
 
-def _autodetect_read_direction(reads, classifier, n=100,
-                               force_read_direction=None):
-    if force_read_direction == 'forward':
+def _autodetect_orientation(reads, classifier, n=100,
+                            force_orientation=None):
+    if force_orientation == 'same':
         return reads
-    if force_read_direction == 'reverse':
+    if force_orientation == 'reverse-complement':
         return (r.reverse_complement() for r in reads)
     reads = iter(reads)
     first_n_reads = list(islice(reads, n))
     result = list(zip(*predict(first_n_reads, classifier, confidence=0.)))
-    _, _, forward_confidence = result
+    _, _, same_confidence = result
     reversed_n_reads = [r.reverse_complement() for r in first_n_reads]
     result = list(zip(*predict(reversed_n_reads, classifier, confidence=0.)))
     _, _, reverse_confidence = result
-    if sum(forward_confidence) > sum(reverse_confidence):
+    if median(array(same_confidence) - array(reverse_confidence)) > 0.:
         return chain(first_n_reads, reads)
     return chain(reversed_n_reads, (r.reverse_complement() for r in reads))
 
@@ -137,10 +138,10 @@ def _autodetect_read_direction(reads, classifier, n=100,
 def classify_sklearn(reads: DNAIterator, classifier: Pipeline,
                      chunk_size: int=262144, n_jobs: int=1,
                      pre_dispatch: str='2*n_jobs', confidence: float=-1.,
-                     force_read_direction: str=None
+                     force_orientation: str=None
                      ) -> pd.DataFrame:
-    reads = _autodetect_read_direction(
-        reads, classifier, force_read_direction=force_read_direction)
+    reads = _autodetect_orientation(
+        reads, classifier, force_orientation=force_orientation)
     predictions = predict(reads, classifier, chunk_size=chunk_size,
                           n_jobs=n_jobs, pre_dispatch=pre_dispatch,
                           confidence=confidence)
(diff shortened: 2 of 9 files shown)


q2_feature_classifier/classifier.py
```diff
@@ -133,20 +156,24 @@ def classify(reads: DNAIterator, classifier: Pipeline,
...
     parameters=_classify_parameters,
     outputs=[('classification', FeatureData[Taxonomy])],
-    name='Classify reads by taxon.',
+    name='Pre-fitted sklearn-based taxonomy classifier',
     description='Classify reads by taxon using a fitted classifier.',
     parameter_descriptions={'confidence': 'Confidence threshold for limiting '
                             'taxonomic depth. Negative value disables. '
-                            'Currently experimental. USE WITH CAUTION.'}
+                            'Currently experimental. USE WITH CAUTION.',
+                            'force_read_direction': 'Force reads to be treated'
```
jairideout: Can you add a brief description of how the autodetection works? Also, this small addition may be useful (or some variation thereof):

*Force reads to be treated as forward or reverse complement with respect to the reference sequences the classifier was fitted to. [...]*

I'm wondering if a parameter name like `read_orientation` would be a little clearer? It could have a default of None for autodetection. "same" would imply the reads are in the same orientation as the reference sequences, and "reverse-complement" would imply the reads need reverse-complementing to be in the same orientation. I'll defer to you and @gregcaporaso for what you think will be clearest to users.

gregcaporaso: I like the parameter name suggestion from @jairideout, I think that will be really clear for users.

q2_feature_classifier/classifier.py
```diff
@@ -104,20 +106,41 @@ def fit_classifier(reference_reads: DNAIterator, reference_taxonomy: pd.Series,
...
+    if force_read_direction == 'reverse':
+        return (r.reverse_complement() for r in reads)
+    reads = iter(reads)
+    first_n_reads = list(islice(reads, n))
+    result = list(zip(*predict(first_n_reads, classifier, confidence=0.)))
+    _, _, forward_confidence = result
+    reversed_n_reads = [r.reverse_complement() for r in first_n_reads]
+    result = list(zip(*predict(reversed_n_reads, classifier, confidence=0.)))
+    _, _, reverse_confidence = result
+    if sum(forward_confidence) > sum(reverse_confidence):
```
gregcaporaso: Would `median` make more sense than `sum` here? I'm just wondering if `sum` could potentially be impacted by outliers.

q2_feature_classifier/_skl.py
```diff
@@ -21,7 +21,8 @@
             'non_negative': True}],
           ['classify',
            {'__type__': 'custom.LowMemoryMultinomialNB',
-            'alpha': 0.01}]]]]
+            'alpha': 0.01,
+            'chunk_size': 20000}]]]]
```
gregcaporaso: It looks like it is (in `q2_feature_classifier/custom.py`), but it would be good to get confirmation on that.

BenKaehler: This parameter could be changed in `custom.py`, or here. I've changed it in both, for peace of mind.

q2_feature_classifier/classifier.py
BenKaehler: I've made the suggested changes. Please check my wording.

q2_feature_classifier/classifier.py
BenKaehler: `sum` is fine because confidence is bounded between zero and one. Also, it has the effect of down-weighting reads with low confidence in either orientation.

... but I've changed it to `median` so that when it breaks it's not my fault :smile:

I've also changed the calculation to take advantage of sample pairing.

jairideout: So what are we ultimately going with here? It sounds like `sum` makes more sense but I don't know which will perform better in the wild. There will always be cases where autodetection fails, but we should pick the most reasonable implementation :P

q2_feature_classifier/_skl.py
jairideout: Specifically, I'm wondering if the error @gregcaporaso ran into (see the linked comment) is no longer raised when using the new default chunk size of 20000. He suggested choosing another chunk size (dynamically) in that case to avoid the error:

> basically we'll just need to detect when 20k isn't a sensible value, and go with something that is.

q2_feature_classifier/classifier.py
```diff
@@ -133,20 +157,27 @@ def classify(reads: DNAIterator, classifier: Pipeline,
...
     outputs=[('classification', FeatureData[Taxonomy])],
-    name='Classify reads by taxon.',
+    name='Pre-fitted sklearn-based taxonomy classifier',
     description='Classify reads by taxon using a fitted classifier.',
     parameter_descriptions={'confidence': 'Confidence threshold for limiting '
-                            'taxonomic depth. Negative value disables. '
-                            'Currently experimental. USE WITH CAUTION.'}
+                            'taxonomic depth. Negative value disables '
+                            'Currently experimental. USE WITH CAUTION',
+                            'force_orientation': 'Force assumption that reads '
```
jairideout: I'm not in love with the name `force_orientation`; the word "force" seems unnecessarily strong here, especially when `"same"` is passed -- there's no "forcing" in that case, the reads are classified as usual without transformation. The default case (autodetection) also makes "forcing" seem out of context here.

Since the user is describing the orientation their input reads w.r.t the reference, what about `read_orientation` (I suggested this [here](https://github.com/qiime2/q2-feature-classifier/pull/62#discussion_r110752081))?

q2_feature_classifier/classifier.py
```diff
@@ -133,20 +157,27 @@ def classify(reads: DNAIterator, classifier: Pipeline,
...
+                            'have the same orientation as the reference '
+                            'sequences or that reads must be reversed and '
```
jairideout: Can you describe what the `"same"` and `"reverse-complemented"` options do in the context of this text?

q2_feature_classifier/classifier.py
jairideout: Thanks! See my followup comments [here](https://github.com/qiime2/q2-feature-classifier/pull/62#discussion_r110963462) and [here](https://github.com/qiime2/q2-feature-classifier/pull/62#discussion_r110963736).

q2_feature_classifier/classifier.py
gregcaporaso: I don't have a good feeling for which will do better, but since confidence is bound by 1.0 (I wasn't thinking about that), maybe sum does make more sense.

q2_feature_classifier/classifier.py
gregcaporaso: I agree with this. @BenKaehler, if you are too tied up to make the change we could do it. Just let us know.

gregcaporaso:
I was able to confirm that the error I mentioned in #57 no longer occurs when I run `qiime feature-classifier fit-classifier-naive-bayes` with this branch. I tested both with using the default for `--p-classify--chunk-size`, and specifying `--p-classify--chunk-size 20000`. 🎉

q2_feature_classifier/_skl.py
gregcaporaso: I was able to confirm that the error I mentioned in #57 no longer occurs when I run `qiime feature-classifier fit-classifier-naive-bayes` with this branch. I tested both with using the default for `--p-classify--chunk-size`, and specifying `--p-classify--chunk-size 20000`. 🎉

BenKaehler: Sorry, I forgot to mention that the error that led to @gregcaporaso suggesting a dynamic chunk size was caused by a bug. I fixed the bug, so we don't need a dynamic chunk size. But you could call it a dynamic chunk size if you want. The last chunk was always intended to contain the remainder of the samples, and if the chunk size is bigger than the number of samples then the first chunk is also the last chunk.

q2_feature_classifier/classifier.py
BenKaehler: meh. I don't think it matters too much. I'm going to leave it because if it ain't broke, don't fix it (again).

q2_feature_classifier/classifier.py
BenKaehler: Sorry, I misread your suggestion. I have made the parameter name less forceful.

q2_feature_classifier/classifier.py
BenKaehler: I'm not exactly sure what you mean here, but I've changed the help text in accordance with my interpretation of your question. Please check the wording again (and feel free to change it if you would like to).

q2_feature_classifier/_skl.py
jairideout: Great, thanks for confirming guys! That all sounds good 👍 💯

q2_feature_classifier/classifier.py
jairideout: Sounds good!

q2_feature_classifier/classifier.py
jairideout: No worries, thanks for making that change @BenKaehler!

q2_feature_classifier/classifier.py
jairideout: Sorry for the poor explanation -- rereading my comment, it makes no sense. Your changes look great, that was what I was looking for :P

jairideout:
Looks great, thanks @BenKaehler!!

gregcaporaso:
Thanks @BenKaehler and @jairideout!

BenKaehler:
Thanks @jairideout and @gregcaporaso (for the second time, I thought the `docs` PR was this one).

PR #68, opened 2017-05-16
ENH: Add default confidence value for classify-sklearn (#64). Also set default naive-bayes prior to uniform.

### LINKED ISSUE #64: add default confidence value > 0.0 for classify-sklearn
The current default (-1.0) doesn't compute confidence scores. This can give very misleading results in some cases. For example, I accidentally used a 515F/806R classifier on data that was sequenced with 27F/338R with the default of no confidence estimation. I got species-level assignments, but they were nowhere near the right taxa. Since there were no confidence estimates, there was no indication that anything was wrong with the results. When I re-ran the data with higher confidence values (0.8 and 0.5) I only had assignments of _bacteria_ (in a few bases, a phylum-level assignment was also made with confidence=0.5). Indicating that they were unassigned would have been ideal, but in this case I could clearly see that there was a problem. I'm labeling this as a bug because there is a very real potential for users to get a silent failure like the one I'm describing here. 

@BenKaehler and @nbokulich are going to make a decision on a default value to start with based on their on-going parameter sweep/optimization. 

Also, when using a value other than -1.0 for confidence, the confidence scores will be included in the `FeatureData[Taxonomy]`'s data. It would be really nice if ``qiime taxa tabulate`` could present these values as well (it could just present all of the information in the `data/taxonomy.tsv` file). This is secondary to the main point of this issue though, which is just starting to use a confidence value > 0.0.

BenKaehler:
Fixes #64.


q2_feature_classifier/_skl.py
```diff
@@ -21,7 +21,8 @@
             'non_negative': True}],
           ['classify',
            {'__type__': 'custom.LowMemoryMultinomialNB',
-            'alpha': 0.01}]]]]
+            'alpha': 0.01,
+            'fit_prior': False}]]]]
```
jairideout: Is this `fit_prior` change something we should notify users about in the release notes/announcement in addition to the change in default confidence? If so, can you provide some text describing what the change is and how it affects users? Then I can add the text to our release notes draft.

BenKaehler: The release notes might be:

- changed default taxon prior probabilities to uniform for taxonomy assignment, rather than estimating them from the reference database. This is in keeping with other standard approaches (including RDP Classifier). Users might notice small changes in classification confidence.

In future, is there a way that I can add to the release notes directly? Thanks!

thermokarst: @BenKaehler:

> In future, is there a way that I can add to the release notes directly? Thanks!

Yep! I will ping you on Slack with directions.

jairideout: Thanks for the release notes text @BenKaehler!

q2_feature_classifier/classifier.py
```diff
@@ -170,8 +170,7 @@ def classify_sklearn(reads: DNAIterator, classifier: Pipeline,
     name='Pre-fitted sklearn-based taxonomy classifier',
     description='Classify reads by taxon using a fitted classifier.',
     parameter_descriptions={'confidence': 'Confidence threshold for limiting '
-                            'taxonomic depth. Negative value disables '
-                            'Currently experimental. USE WITH CAUTION',
+                            'taxonomic depth. Negative value disables.',
```
gregcaporaso: Can you can this to: 

```
'taxonomic depth. Provide -1 to disable '
'confidence calculation, or 0 to calculate '
'confidence but not apply it to limit the '
'taxonomic depth of the assignments.'
```

I know that you can provide other negative values, but I think we'll get fewer support questions about this if we are just very specific about what the value should be.

gregcaporaso:
@BenKaehler, just one change request on this. We can ignore the decrease in test coverage, it's not a meaningful change.

gregcaporaso:
@BenKaehler approved the change I proposed above, but isn't available to make it this week. I'm going to make the change so we can get this one merge.

Issue #65, opened 2017-05-10
Add 5' trimming to extract-reads


BenKaehler:
It is sometimes practice to trim reads at both ends. We should support that in `extract-reads`.

Thanks @nbokulich for bringing this to my attention.

BenKaehler:
From @nbokulich:

we could use the same parameter names as dada2 does: `trim_left` and `trunc_len`

the read is cut to `trunc_len`, then the first `trim_left` nucleotides are removed

PR #69, opened 2017-05-17
ENH: Add 5' trimming to extract-reads (#65)

BenKaehler:
Also added some parameter help to `extract-reads`.


q2_feature_classifier/tests/test_cutter.py
```diff
@@ -23,16 +23,18 @@ def setUp(self):
 
     def test_extract_reads(self):
         # extract_reads should generate a FeatureData(Sequence) full of reads
-        length = 75
+        trunc_len = 75
```
thermokarst: Can you add some additional tests to cover the [branches introduced](https://github.com/qiime2/q2-feature-classifier/pull/69/files#diff-543f7c49a71dc85a194d9f5fa66946c8R115) in the PR? Basically checking output when `trunc_len` is negative, `trim_left` is negative, and both params are negative (ie not supplied). Those could be new tests, or you could do some kind of loop in this test, I don't have a preference.

q2_feature_classifier/_cutter.py
```diff
@@ -152,11 +161,21 @@ def extract_reads(sequences: DNAIterator,  f_primer: str, r_primer: str,
...
+    parameters={'trunc_len': Int,
+                'trim_left': Int,
                 'f_primer': Str,
                 'r_primer': Str,
                 'identity': Float},
     outputs=[('reads', FeatureData[Sequence])],
     name='Extract reads from reference',
-    description='Extract sequencing-like reads from a reference database.'
+    description='Extract sequencing-like reads from a reference database.',
+    parameter_descriptions={'f_primer': 'forward primer sequence',
```
thermokarst: Thank you! Adding `parameter_descriptions` is on our todo list!

q2_feature_classifier/_cutter.py
```diff
@@ -109,17 +108,22 @@ def _gen_reads(sequences: DNAIterator,  f_primer: str, r_primer: str,
...
         if not amp:
             continue
-        if length > 0:
-            amp = amp[:length]
         yield amp
 
 
 def extract_reads(sequences: DNAIterator,  f_primer: str, r_primer: str,
-                  length: int=-1, identity: float=0.8) -> DNAIterator:
+                  trunc_len: int=-1, trim_left: int=-1,
```
thermokarst: I don't have a preference on the `-1` sentinels, but I think @ebolyen and/or @jairideout should chime in, in case they have some thoughts.

ebolyen: Is `0` a useful value? I think it could make sense as the sentinel value. Otherwise this seems fine.

I think a perfect solution would be:
`trunc_len: Union[str, int]='None'` 
with the primitive type being:
 `Int % Range(0, None) | Str % Choices({'None'})` 
but we don't support that at the moment.

BenKaehler:
Thanks Everyone, I think my changes addressed your comments. Please let me know if I missed anything.

q2_feature_classifier/tests/test_cutter.py
```diff
@@ -18,23 +22,43 @@ class CutterTests(FeatureClassifierTestPluginBase):
...
-        self.assertTrue(bool(outseqs))
-        for seq in outseqs:
-            self.assertEqual(len(seq), trunc_len - trim_left)
+
+        trunc_lens = 0, 75, 5
+        trim_lefts = 0, 5
+        raw_lens = []
+        for trunc_len, trim_left in product(trunc_lens, trim_lefts):
+            if trunc_len == trim_left and trunc_len > 0:
+                with self.assertRaises(RuntimeError):
```
thermokarst: Can you make this an `assertRaisesRegex` --- they are @jairideout's favorite and i think he would be excited to see one here:

```python
with self.assertRaisesRegex(RuntimeError, "No matches found"):
```

q2_feature_classifier/tests/test_cutter.py
```diff
@@ -18,23 +22,43 @@ class CutterTests(FeatureClassifierTestPluginBase):
 
     def setUp(self):
         super().setUp()
-        self.sequences = Artifact.import_data(
-            'FeatureData[Sequence]', self.get_data_path('dna-sequences.fasta'))
+        seqs = skbio.io.read(self.get_data_path('dna-sequences.fasta'),
+                             format='fasta', constructor=skbio.DNA)
+        tmpseqs = os.path.join(self.temp_dir.name, 'temp-seqs.fasta')
+        skbio.io.write((s for s in islice(seqs, 10)), 'fasta', tmpseqs)
```
thermokarst: Wow, the tests are so fast now!!

q2_feature_classifier/tests/test_cutter.py
```diff
@@ -18,23 +22,43 @@ class CutterTests(FeatureClassifierTestPluginBase):
...
-        outseqs = list(result.reads.view(DNAIterator))
-        self.assertLessEqual(len(outseqs), len(inseqs))
-        self.assertTrue(bool(outseqs))
-        for seq in outseqs:
-            self.assertEqual(len(seq), trunc_len - trim_left)
+
+        trunc_lens = 0, 75, 5
+        trim_lefts = 0, 5
+        raw_lens = []
+        for trunc_len, trim_left in product(trunc_lens, trim_lefts):
```
thermokarst: 👍

thermokarst:
Looks great! Thanks @BenKaehler!! 🎉 🎉

Issue #67, opened 2017-05-16
unassigned sequences


BenKaehler:
It would be clearer for the users if, when the classifier failed to classify at the confidence threshold, it outputted a null classification, rather than omitting the sequence from the results. So that's what we'll do.

Thanks @nbokulich for suggesting this.

thermokarst:
Fixed in #67.

PR #70, opened 2017-05-17
ENH: unassigned sequences (#67)

BenKaehler:
`classify-sklearn` now returns Unassigned if confidence does not meet threshold, instead of returning nothing for that sequence.


maxvonhippel:
This makes sense to me.  I ran the tests with `pytest` locally, and ran the commands from the tutorials in the docs, and everything worked as expected.

q2_feature_classifier/tests/test_classifier.py
```diff
@@ -29,36 +30,29 @@ def setUp(self):
         self.taxonomy = Artifact.import_data(
             'FeatureData[Taxonomy]', self.get_data_path('taxonomy.tsv'))
 
+        seq_path = self.get_data_path('se-dna-sequences.fasta')
+        reads = Artifact.import_data('FeatureData[Sequence]', seq_path)
+        fitter_name = _specific_fitters[0][0]
+        fitter = getattr(feature_classifier.methods,
+                         'fit_classifier_' + fitter_name)
+        self.classifier = fitter(reads, self.taxonomy).classifier
```
ebolyen: :+1:

q2_feature_classifier/tests/test_classifier.py
```diff
@@ -102,31 +96,41 @@ def test_classify(self):
...
+
+    def test_unassigned_taxa(self):
+        # classifications that don't meet the threshold should be "Unassigned"
+        classify = feature_classifier.methods.classify_sklearn
+        seq_path = self.get_data_path('se-dna-sequences.fasta')
+        reads = Artifact.import_data('FeatureData[Sequence]', seq_path)
+        result = classify(reads, self.classifier, confidence=1.)
+
+        ref = self.taxonomy.view(pd.Series).to_dict()
+        cls = result.classification.view(pd.Series).to_dict()
```
ebolyen: Could we use a different name here? I can't help but read this as `class` and then I see `cls[_id]` and start thinking we're dealing with metaclasses somehow...

q2_feature_classifier/tests/test_classifier.py
```diff
@@ -102,31 +96,41 @@ def test_classify(self):
...
+
+        assert 'Unassigned' in cls.values()
+        for seq in reads.view(DNAIterator):
+            _id = seq.metadata['id']
```
ebolyen: Oops, one more comment:
`_id` -> `id_`
stylistic again, but when you're avoiding clobbering defaults the underscore usually comes after, because it doesn't imply anything about information hiding.

BenKaehler:
Thanks Everyone, I think the last push fixes up those concerns.

PR #75, opened 2017-05-29
FIX: fixes #74 enough

### LINKED ISSUE #74: fit-classifier-naive-bayes raises a ValueError for valid input
The problem:
```
$ qiime feature-classifier fit-classifier-naive-bayes --p-feat-ext--n-features 1024 --p-feat-ext--ngram-range '[4,16]' --i-reference-reads ref_dbs/99_gg_seq.qza --i-reference-taxonomy ref_dbs/99_gg_tax.qza --o-classifier never_gonna_get_that_far.qza
Plugin error from feature-classifier:

  indices and data should have the same size

Re-run with --verbose to see debug info.
```

This is a manifestation of a bug in scikit-learn: scikit-learn/scikit-learn#8941. 

It could be fixed by kludging the feature classifier to chunk the input to HashingVectorizer. It could be fixed by fixing scikit-learn. Not sure what the best approach is yet.

BenKaehler:
Added `ChunkedHashingVectorizer` to the `custom` module, which will get anyone who has this problem out of trouble.

Have not made `ChunkedHashingVectorizer` the default because under ordinary circumstances it makes no difference, other than adding another potential point of failure.


q2_feature_classifier/custom.py
```diff
@@ -39,6 +43,46 @@ def fit(self, X, y, sample_weight=None):
         return self
 
 
+class ChunkedHashingVectorizer(HashingVectorizer):
+    # This class is a kludge to get around
+    # https://github.com/scikit-learn/scikit-learn/issues/8941
```
thermokarst: 👍 Thanks for the reminder!

Issue #84, opened 2017-08-18
ChunkedHashingVectorizer is not compatible with sklearn 0.19.0


ebolyen:
[This testing log](https://busywork.qiime2.org/teams/main/pipelines/master-distribution/jobs/unit-test-q2-feature-classifier/builds/65) shows the issue. It looks like a parameter was removed (or perhaps it became keyword only?). In any case, a `super().transform` method has a different signature now.

thermokarst:
Hey @BenKaehler, do you have time to look at this for this release (2017.8) or the next release (2017.9)? If we tried to get this in for 2017.8, a PR would need to get in this week some time (we are shooting for a release on August 30th). No worries if you don't have the bandwidth to look into this right now, please just let us know either way. Thanks!

BenKaehler:
Thanks Matt, I'll try to get this done this week.

PR #86, opened 2017-08-24
FIX: updated for changes to HashingVectorizer in scikit-learn 0.19 and 0.21 (#84)

BenKaehler:
Adapts to changes to `sklearn.feature_extraction.text.HashingVectorizer.transform` in `scikit-learn` 0.19.0 and preempts deprecation of `non_negative` parameter in the same in `scikit-learn` 0.21.0.


thermokarst:
@BenKaehler: dependency changes are something that [busywork](https://busywork.qiime2.org) doesn't handle super well right now, wrt Travis. I think for this PR we will have to test locally, and ignore Travis results. Then, busywork should be able to pick up on the new changes and flush out the [environment files](https://github.com/qiime2/environment-files) with the new version of sklearn. Thanks!

PR #98, opened 2017-12-07
reads per batch autotuner

### LINKED ISSUE #90: Should classify-sklearn error when "dummy" jobs are created?
When running `classify-sklearn`, it's up to the user to do the load-balancing per job such that the parallelism works as best as possible. As such when running `N` (N > 1) parallel jobs, there might be situations where the number of reads to be processed is smaller than the number of reads that each job can process, and as such all reads are processed by one job, still all other `N-1` jobs are created (note that each job will require some space to allocate the pre-trained classifier, which can be a lot for big databases like Silva). This gives the untrained eye the impression that parallelism is or isn't working.

An easy solution would to check before the classification starts, and produce an error with a proposed solution. An alternative would be to not create the `N-1` jobs that will do nothing (though this would probably be prone to misinterpretations).

nbokulich:
fixes #90 

This is the solution proposed by @BenKaehler in #90 (if `n_jobs` != 1, set `reads-per-batch` = number of reads / n_jobs), not the solution originally proposed by @ElDeveloper (raise an error suggesting a better setting).

The autotuner is enabled if `reads_per_batch` = 0 (new default), but otherwise users can manually set this value (e.g., to disable the autotuner) so I have not removed this parameter.

The autotuner has a negligible impact on runtime:

image

### DIFF
--- a/q2_feature_classifier/classifier.py
+++ b/q2_feature_classifier/classifier.py
@@ -20,7 +20,7 @@ from q2_types.feature_data import (
 from q2_types.feature_table import FeatureTable, RelativeFrequency
 from sklearn.pipeline import Pipeline
 import sklearn
-from numpy import median, array
+from numpy import median, array, ceil
 import biom
 import skbio
 
@@ -177,13 +177,18 @@ def _autodetect_orientation(reads, classifier, n=100,
 
 def _autotune_reads_per_batch(reads, n_jobs):
     # detect effective jobs. Will raise error if n_jobs == 0
-    n_jobs = sklearn.externals.joblib.effective_n_jobs(n_jobs)
+    if n_jobs == 0:
+        raise ValueError("Value other than zero must be specified as number "
+                         "of jobs to run.")
+    else:
+        n_jobs = sklearn.externals.joblib.effective_n_jobs(n_jobs)
+
     # we really only want to calculate this if running in parallel
     if n_jobs != 1:
         seq_count = subprocess.run(
             ['grep', '-c', '^>', str(reads)], check=True,
             stdout=subprocess.PIPE)
-        return int(int(seq_count.stdout.decode('utf-8')) / n_jobs)
+        return int(ceil(int(seq_count.stdout.decode('utf-8')) / n_jobs))
     # otherwise reads_per_batch = large value (similar to original default)
     else:
         return 500000
--- a/q2_feature_classifier/classifier.py
+++ b/q2_feature_classifier/classifier.py
@@ -11,15 +11,18 @@ import importlib
 import inspect
 import warnings
 from itertools import chain, islice
+import subprocess
 
 import pandas as pd
-from qiime2.plugin import Int, Str, Float, Bool, Choices
-from q2_types.feature_data import FeatureData, Taxonomy, Sequence, DNAIterator
+from qiime2.plugin import Int, Str, Float, Bool, Choices, Range
+from q2_types.feature_data import (
+    FeatureData, Taxonomy, Sequence, DNAIterator, DNAFASTAFormat)
 from q2_types.feature_table import FeatureTable, RelativeFrequency
 from sklearn.pipeline import Pipeline
 import sklearn
 from numpy import median, array
 import biom
+import skbio
 
 from ._skl import fit_pipeline, predict, _specific_fitters
 from ._taxonomic_classifier import TaxonomicClassifier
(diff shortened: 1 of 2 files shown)


gregcaporaso:
> The autotuner has a negligible impact on runtime

@nbokulich, I think you mean that the overhead of running the autotuner itself is negligible, but the effect of auto-tuning (versus the previous behavior of just using a very large default value) is a considerable decrease in run time. Is that right?

nbokulich:
Thanks @gregcaporaso ! Yes and yes. See plot below.

> I think you mean that the overhead of running the autotuner itself is negligible

Correct

> the effect of auto-tuning (versus the previous behavior of just using a very large default value) is a considerable decrease in run time. Is that right?

Correct. The previous plot I posted was user time. Here's real time, with 4 jobs. Note how setting a very high `reads-per-batch` (10000) takes nearly 4X as long, because only 1 job is effectively performed:

image

q2_feature_classifier/classifier.py
```diff
@@ -172,11 +175,33 @@ def _autodetect_orientation(reads, classifier, n=100,
...
-                     reads_per_batch: int=262144, n_jobs: int=1,
+def _autotune_reads_per_batch(reads, n_jobs):
+    # detect effective jobs. Will raise error if n_jobs == 0
+    n_jobs = sklearn.externals.joblib.effective_n_jobs(n_jobs)
+    # we really only want to calculate this if running in parallel
+    if n_jobs != 1:
+        seq_count = subprocess.run(
+            ['grep', '-c', '^>', str(reads)], check=True,
+            stdout=subprocess.PIPE)
+        return int(int(seq_count.stdout.decode('utf-8')) / n_jobs)
```
gregcaporaso: I don't think is computation is exactly right as it can underestimate the number of reads per batch. 

For example, if the sequence count is 11, and `n_jobs` = 5, then the reads per batch as a float will be 2.2. Casting that to `int` will give 2 reads per batch, so I think ultimately 6 jobs would need to be run. That could really slow things down, I think (since I'm guessing it would wait for one job to complete before starting up a new one?). 

You could get around this by calling `ceil` before `int`, as that would overestimate the number of reads per job. For example:

```python
>>> import numpy as np

>>> int(11/5)
2

>>> int(np.ceil(11/5))
3
```

q2_feature_classifier/classifier.py
```diff
@@ -172,11 +175,33 @@ def _autodetect_orientation(reads, classifier, n=100,
...
+def classify_sklearn(reads: DNAFASTAFormat, classifier: Pipeline,
+                     reads_per_batch: int=0, n_jobs: int=1,
                      pre_dispatch: str='2*n_jobs', confidence: float=0.7,
                      read_orientation: str=None
                      ) -> pd.DataFrame:
+    # autotune reads per batch
+    if reads_per_batch == 0:
+        reads_per_batch = _autotune_reads_per_batch(reads, n_jobs)
+
+    # transform reads to DNAIterator
```
gregcaporaso: Is this still needed here, since you're calling `grep` to count the reads?

q2_feature_classifier/tests/test_classifier.py
```diff
@@ -223,3 +223,16 @@ def test_unassigned_taxa(self):
             id_ = seq.metadata['id']
             assert ref[id_].startswith(classified[id_]) or \
                 classified[id_] == 'Unassigned'
+
+    def test_autotune_reads_per_batch(self):
+        self.assertEqual(
```
gregcaporaso: Could you add tests including a few different values of `n_jobs` in here, ideally including one that would catch the `int(ceil())` versus `int` case I mentioned above?

gregcaporaso: Also, what happens if `n_jobs` is greater than the number of reads? That should be covered in a test (`ceil` might handle that correctly - I think we'd want the reads per batch to be 1 in this edge case).

q2_feature_classifier/tests/test_classifier.py
```diff
@@ -223,3 +223,16 @@ def test_unassigned_taxa(self):
...
+            _autotune_reads_per_batch(self.seq_path, n_jobs=4), 276)
+
+    def test_autotune_reads_per_batch_disable_if_single_job(self):
+        self.assertEqual(
+            _autotune_reads_per_batch(self.seq_path, n_jobs=1), 500000)
+
+    def test_autotune_reads_per_batch_zero_jobs(self):
```
gregcaporaso: Could you catch this case yourself and raise your own error? I think our error would be better. For example:

```python
if n_jobs == 0:
    raise ValueError("Value other than zero must be specified as number of jobs to run.")
```

q2_feature_classifier/classifier.py
nbokulich: Ben originally set up `classify_sklearn` to accept a `DNAIterator` as input — so I added this step to avoid disturbing anything downstream. This seemed more straightforward and took less code than creating a copy of `reads` and transforming into `DNAFASTAFormat`. Am I missing something, though? e.g., can I get a path view from a `DNAIterator`?

nbokulich:
Thanks @gregcaporaso ! I have made all suggested changes (but pls see my response about `DNAIterator` above)

q2_feature_classifier/classifier.py
gregcaporaso: Sorry, I didn't catch that the view type had changed in this PR. This is good as-is.

gregcaporaso:
I did some tests with real data, and it all seems to be working as expected. Thanks @nbokulich!

PR #117, opened 2018-11-01
fixed vsearch maxaccepts range

### LINKED ISSUE #114: `classify-consensus-vsearch`: allow maxaccepts=0
**Bug Description**
The help docs state that maxaccepts should be in the range 0 to infinity, but using 0 raises an error:
```
Error: Invalid value for "--p-maxaccepts": 0 is smaller than the minimum valid value 1.
```

**Steps to reproduce the behavior**
set maxaccepts=0

**Expected behavior**
We should allow this setting — it would match against all sequences in the database to find consensus. Honestly I'm not sure how useful that will be in the context of taxonomy classification, but it will better expose the underlying functionality.

Note the workaround is negligible — if you know the number of seqs in the database, just set that number (or a really high number)

**References**
[forum xref](https://forum.qiime2.org/t/finding-incorrectly-formatted-needle-in-apparently-mostly-formatted-properly-haystack/6493/7?u=nicholas_bokulich)

nbokulich:
fixes #114

### DIFF
--- a/q2_feature_classifier/_blast.py
+++ b/q2_feature_classifier/_blast.py
@@ -57,7 +57,7 @@ plugin.methods.register_function(
         'strand': ('Align against reference sequences in forward ("plus"), '
                    'reverse ("minus"), or both directions ("both").'),
         'maxaccepts': ('Maximum number of hits to keep for each query. Must '
-                       'be in range [0, infinity].'),
+                       'be in range [1, infinity].'),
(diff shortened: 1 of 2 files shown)


thermokarst:
@ChrisKeefe, can you please review this PR?

nbokulich:
thanks @ChrisKeefe ! resolved the merge conflict.

PR #120, opened 2018-12-01
exposed query_cov params

### LINKED ISSUE #100: `classify-consensus-blast` should have a cutoff for query coverage
[forum xref](https://forum.qiime2.org/t/feature-classifier-classify-consensus-blast-should-have-a-cutoff-for-query-cover/3641?u=nicholas_bokulich)

similar to what we have for q2-quality-control [exclude-seqs](https://github.com/qiime2/q2-quality-control/blob/master/q2_quality_control/quality_control.py#L32), see `perc_query_aligned` parameter.

nbokulich:
Fixes #100 

Have exposed parameters to filter alignment results by query coverage, for both vsearch and blast (used qcov_hsp_perc parameter). This is just exposing the underlying parameters for the alignment tools, no new native code, so I have not added unit tests (similarly we do not have unit tests to validate other parameters).


q2_feature_classifier/_vsearch.py
```diff
@@ -19,17 +19,19 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
                                reference_reads: DNAFASTAFormat,
                                reference_taxonomy: pd.Series,
                                maxaccepts: int = 10,
-                               perc_identity: int = 0.8, strand: str = 'both',
+                               perc_identity: int = 0.8,
+                               query_cov: float = 0.8,
```
ChrisKeefe: Looks like you inherited some weirdness here. Should this be `perc_identity: float = 0.8?`

nbokulich: thank you for catching that @ChrisKeefe ! Totally bizarre.

nbokulich: I just fixed that... let me know if other changes are needed.

thermokarst:
Thanks for the review, @ChrisKeefe!

@nbokulich - can you update the changelog?

PR #133, opened 2019-04-25
ENH: `classify-consensus-vsearch` upgrades, plus hybrid classifier

nbokulich:
a hybrid beast is born.

The hybrid classifier is really just a pipeline of `classify-consensus-vsearch` + `classify-sklearn` with some extra bells and whistles thrown in.

This PR also contains some awesome enhancements to `classify-consensus-vsearch`, allowing for exact hit matching and top hit selection.

The hybrid classifier works for me but could use some more alpha testing if anyone is interested. Or we can unleash and see what chaos ensues.

### DIFF
--- a/q2_feature_classifier/_consensus_assignment.py
+++ b/q2_feature_classifier/_consensus_assignment.py
@@ -40,7 +40,7 @@ def _consensus_assignments(
             consensus = {'': ('', '')}
         result = pd.DataFrame.from_dict(consensus, 'index')
         result.index.name = 'Feature ID'
-        result.columns = ['Taxon', 'Confidence']
+        result.columns = ['Taxon', 'Consensus']
         return result
 
 
--- a/q2_feature_classifier/_consensus_assignment.py
+++ b/q2_feature_classifier/_consensus_assignment.py
@@ -11,6 +11,7 @@ import subprocess
 import pandas as pd
 from os.path import isfile
 from collections import Counter, defaultdict
+import qiime2
 
 
 def _get_default_unassignable_label():
@@ -242,3 +243,9 @@ def _compute_consensus_annotation(annotations, min_consensus,
         consensus_fraction_result = 1.0
 
     return annotation, consensus_fraction_result
+
+
+def _annotate_method(taxa, method):
+    taxa = taxa.view(pd.DataFrame)
+    taxa['method'] = method
+    return qiime2.Artifact.import_data('FeatureData[Taxonomy]', taxa)
--- a/q2_feature_classifier/_vsearch.py
+++ b/q2_feature_classifier/_vsearch.py
@@ -36,6 +36,8 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
                                threads: str = 1) -> pd.DataFrame:
     seqs_fp = str(query)
     ref_fp = str(reference_reads)
+    if maxaccepts == 'all':
+        maxaccepts = 0
     cmd = ['vsearch', '--usearch_global', seqs_fp, '--id', str(perc_identity),
            '--query_cov', str(query_cov), '--strand', strand, '--maxaccepts',
            str(maxaccepts), '--maxrejects', '0', '--output_no_hits', '--db',
@@ -143,7 +145,7 @@ parameter_descriptions = {
     'strand': 'Align against reference sequences in forward ("plus") '
               'or both directions ("both").',
     'maxaccepts': 'Maximum number of hits to keep for each query. Set to '
-                  '0 to keep all hits > perc_identity similarity. Must '
+                  '"all" to keep all hits > perc_identity similarity. Must '
                   'be in range [0, infinity].',
     'perc_identity': 'Reject match if percent identity to query is '
                      'lower. Must be in range [0.0, 1.0].',
--- a/q2_feature_classifier/_vsearch.py
+++ b/q2_feature_classifier/_vsearch.py
@@ -15,7 +15,8 @@ from q2_types.feature_data import (
 from .plugin_setup import plugin, citations
 from qiime2.plugin import Int, Str, Float, Choices, Range, Bool
 from ._consensus_assignment import (_consensus_assignments, _run_command,
-                                    _get_default_unassignable_label)
+                                    _get_default_unassignable_label,
+                                    _annotate_method)
 from ._taxonomic_classifier import TaxonomicClassifier
 
 
@@ -92,9 +93,18 @@ def classify_hybrid_vsearch_sklearn(ctx,
                  strand=strand, min_consensus=min_consensus,
                  search_exact=True, threads=threads)
 
+    # Annotate taxonomic assignments with classification method
+    taxa1 = _annotate_method(taxa1, 'VSEARCH')
+
+    # perform second pass classification on unassigned taxa
     # filter out unassigned seqs
-    query, = filter_seqs(sequences=query, taxonomy=taxa1,
-                         include=_get_default_unassignable_label())
+    try:
+        query, = filter_seqs(sequences=query, taxonomy=taxa1,
+                             include=_get_default_unassignable_label())
+    except ValueError:
+        # get ValueError if all sequences are filtered out.
+        # so if no sequences are unassigned, return exact match results
+        return taxa1
 
     # classify with sklearn classifier
     taxa2, = cs(reads=query, classifier=classifier,
@@ -102,11 +112,6 @@ def classify_hybrid_vsearch_sklearn(ctx,
                 confidence=confidence, read_orientation=read_orientation)
 
     # Annotate taxonomic assignments with classification method
-    def _annotate_method(taxa, method):
-        taxa = taxa.view(pd.DataFrame)
-        taxa['method'] = method
-        return qiime2.Artifact.import_data('FeatureData[Taxonomy]', taxa)
-    taxa1 = _annotate_method(taxa1, 'VSEARCH')
     taxa2 = _annotate_method(taxa2, 'sklearn')
 
     # merge into one big happy result
--- a/q2_feature_classifier/_vsearch.py
+++ b/q2_feature_classifier/_vsearch.py
@@ -131,7 +131,7 @@ parameters = {'maxaccepts': Int % Range(1, None) | Str % Choices(['all']),
               'strand': Str % Choices(['both', 'plus']),
               'min_consensus': Float % Range(0.5, 1.0, inclusive_end=True,
                                              inclusive_start=False),
-              'threads': Int}
+              'threads': Int % Range(1, None)}
 
 inputs = {'query': FeatureData[Sequence],
           'reference_reads': FeatureData[Sequence],
--- a/q2_feature_classifier/_vsearch.py
+++ b/q2_feature_classifier/_vsearch.py
@@ -145,15 +145,13 @@ parameter_descriptions = {
     'strand': 'Align against reference sequences in forward ("plus") '
               'or both directions ("both").',
     'maxaccepts': 'Maximum number of hits to keep for each query. Set to '
-                  '"all" to keep all hits > perc_identity similarity. Must '
-                  'be in range [0, infinity].',
+                  '"all" to keep all hits > perc_identity similarity.',
     'perc_identity': 'Reject match if percent identity to query is '
-                     'lower. Must be in range [0.0, 1.0].',
+                     'lower.',
     'query_cov': 'Reject match if query alignment coverage per high-'
-                 'scoring pair is lower. Must be in range [0.0, 1.0].',
+                 'scoring pair is lower.',
     'min_consensus': 'Minimum fraction of assignments must match top '
-                     'hit to be accepted as consensus assignment. Must '
-                     'be in range (0.5, 1.0].',
+                     'hit to be accepted as consensus assignment.',
     'threads': 'Number of threads to use for job parallelization.'}
 
 outputs = [('classification', FeatureData[Taxonomy])]
@@ -221,8 +219,8 @@ plugin.pipelines.register_function(
                             'orientation based on the confidence estimates '
                             'for the first 100 reads.',
         'reads_per_batch': 'Number of reads to process in each batch for '
-                           'sklearn classification. If 0, this parameter is '
-                           'autoscaled to min(number of query sequences / '
+                           'sklearn classification. If "auto", this parameter '
+                           'is autoscaled to min(number of query sequences / '
                            'threads, 20000).',
         'prefilter': 'Toggle positive filter of query sequences on or off.'
     },
--- a/q2_feature_classifier/_vsearch.py
+++ b/q2_feature_classifier/_vsearch.py
@@ -94,15 +94,23 @@ def classify_hybrid_vsearch_sklearn(ctx,
 
     # filter out unassigned seqs
     query, = filter_seqs(sequences=query, taxonomy=taxa1,
-                         exclude=_get_default_unassignable_label())
+                         include=_get_default_unassignable_label())
 
     # classify with sklearn classifier
     taxa2, = cs(reads=query, classifier=classifier,
                 reads_per_batch=reads_per_batch, n_jobs=threads,
                 confidence=confidence, read_orientation=read_orientation)
 
+    # Annotate taxonomic assignments with classification method
+    def _annotate_method(taxa, method):
+        taxa = taxa.view(pd.DataFrame)
+        taxa['method'] = method
+        return qiime2.Artifact.import_data('FeatureData[Taxonomy]', taxa)
+    taxa1 = _annotate_method(taxa1, 'VSEARCH')
+    taxa1 = _annotate_method(taxa2, 'sklearn')
+
     # merge into one big happy result
-    taxa, = merge(data=[taxa1, taxa2])
+    taxa, = merge(data=[taxa2, taxa1])
     return taxa
 
 
--- a/q2_feature_classifier/_vsearch.py
+++ b/q2_feature_classifier/_vsearch.py
@@ -107,7 +107,7 @@ def classify_hybrid_vsearch_sklearn(ctx,
         taxa['method'] = method
         return qiime2.Artifact.import_data('FeatureData[Taxonomy]', taxa)
     taxa1 = _annotate_method(taxa1, 'VSEARCH')
-    taxa1 = _annotate_method(taxa2, 'sklearn')
+    taxa2 = _annotate_method(taxa2, 'sklearn')
 
     # merge into one big happy result
     taxa, = merge(data=[taxa2, taxa1])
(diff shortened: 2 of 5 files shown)


BenKaehler:
Without making any judgements regarding whether the hybrid classifier is a good idea, I have run it and the results are a bit unpredictable.

If the input file contains no or some exact matches I get two-column output with no confidence output or classifier column, regardless of the `prefilter` setting.

If it only uses the vsearch classifier it seems to be ok.

A minor issue is that the help for `reads-per-batch` contains a reference to `n_jobs`, which should be `threads`, I think.

It might be worth mentioning that `confidence` should be interpreted differently depending on which classifier contributed the classification.

nbokulich:
thanks @BenKaehler ! One thing to note:

> If the input file contains no or some exact matches I get two-column output with no confidence output or classifier column, regardless of the prefilter setting.
If it only uses the vsearch classifier it seems to be ok.

This is actually expected, due to the way `merge_taxa` works. I fixed that behavior in this PR: https://github.com/qiime2/q2-feature-table/pull/211

and once that is merged I can add a test for that here (I have confirmed manually that it works a few different ways with this classifier).

Thanks for catching the text changes! Will push those changes in a second.

thermokarst:
As you can tell @nbokulich, we are all chomping at the bit when it comes to reviewing this... Sorry...

ebolyen:
> It might be worth mentioning that confidence should be interpreted differently depending on which classifier contributed the classification.

Forgive me if it already does this, but should there be a second "metadata" column in the artifact to indicate which classifier was used?

devonorourke:
Hey folks,

I noticed this thread while looking for the underlying VSEARCH classifier code. Just wanted to let you know (as @nbokulich already does) that I've been toying around with this idea of using VSEARCH and Naive Bayes in combination for a while with my bat guano COI amplicon datasets.  So far, I've tested each separately, and am working out what kinds of parameters might be useful in combination.

I wanted to point out a couple of things that are 100% anecdote, thinking that maybe your internal testing is seeing similar things.

1. I've tested a very small mock community with these classifiers, but the mock community I was given suffered from a huge problem in that they didn't use voucher specimens to assign taxonomy. Thus, when I say that Naive Bayes tends to "_over classify_" this could be because the existing undefined taxonomies are actually incomplete (the 'truth' in the eyes of the taxonomists only goes to Family or Genus level for some taxa). It might also be because the program is overly aggressive in assigning species names though. I don't know and I think what would be useful for me to do is create a synthetic mock community reflective of a broader arthropod COI dataset. You can see those results [here](https://github.com/devonorourke/tidybug/blob/master/figures/cl_0_mockGroundTruth.png).

I also tested these same classifiers using a bigger real dataset consisting of COI amplicons generated from bat guano. You can see [here](https://github.com/devonorourke/tidybug/blob/master/figures/cl_1_guano_FracDataPerTaxon.png) that the Naive Bayes classifier tends to assign information to Phylum/Class to just about every ASV, but it tends not to assign too many more Genus/Species names than other classifiers (when those other classifiers do not assign a name at that same level). If you're wondering "_what would a giant Venn diagram look like in an [upset plot](https://caleydo.org/tools/upset/) style instead?_" well I have that for you too: [see here](https://github.com/devonorourke/tidybug/blob/master/figures/cl_2_guano_upset.png). What you'll notice is that by and large most things are generally in agreement, but:
a) SINTAX/Naive Bayes tend to assign more things at higher levels
b) SINTAX/Naive Bayes tend to agree more with each other than with alignment approaches 

Question for you: rather than creating new biological mock samples, do you have any suggestions on how to go about creating synthetic mock communities from existing references? I can search through a few databases where supposedly vouchered specimen exist so I can test again with full species named sequences.

nbokulich:
> Question for you: rather than creating new biological mock samples, do you have any suggestions on how to go about creating synthetic mock communities from existing references? I can search through a few databases where supposedly vouchered specimen exist so I can test again with full species named sequences.

Check out tax-credit, you can use cross-validation to test classification of vouchered reference sequences

nbokulich:
>Forgive me if it already does this, but should there be a second "metadata" column in the artifact to indicate which classifier was used?

You are forgiven, @ebolyen : this is already performed, see the `_annotate_method` function that was added as part of this PR.

q2_feature_classifier/_vsearch.py
```diff
@@ -25,53 +30,159 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
...
+
+    # merge into one big happy result
+    taxa, = merge(data=[taxa2, taxa1])
+    return taxa
+
+
+output_descriptions = {
+    'classification': 'The resulting taxonomy classifications.'}
+
+parameters = {'maxaccepts': Int % Range(0, None),
```
thermokarst: What do you think about using something like this, instead?

```suggestion
parameters = {'maxaccepts': Int % Range(1, None) | Str % Choices(['all']),
```
I imagine you would need to accommodate this in the pipeline, but, I think this is a clearer API for users.

q2_feature_classifier/_vsearch.py
```diff
@@ -81,3 +192,57 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
                  'choosing the top N hits, not the first N hits.'),
     citations=[citations['rognes2016vsearch']]
 )
+
+
+plugin.pipelines.register_function(
+    function=classify_hybrid_vsearch_sklearn,
+    inputs={**inputs, 'classifier': TaxonomicClassifier},
+    parameters={**parameters,
+                'reads_per_batch': Int % Range(0, None),
```
```suggestion
                'reads_per_batch': Int % Range(1, None) | Str % Choices(['auto']),
```

q2_feature_classifier/_vsearch.py
```diff
@@ -81,3 +192,57 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
...
+                            'sequences in pre-trained sklearn classifier. '
+                            'same will cause reads to be classified unchanged'
+                            '; reverse-complement will cause reads to be '
+                            'reversed and complemented prior to '
+                            'classification. Default is to autodetect based '
+                            'on the confidence estimates for the first 100 '
+                            'reads.',
+        'reads_per_batch': 'Number of reads to process in each batch for '
+                           'sklearn classification. If 0, this parameter is '
+                           'autoscaled to min( number of query sequences / '
```
```suggestion
                           'autoscaled to min(number of query sequences / '
```

q2_feature_classifier/_vsearch.py
```diff
@@ -81,3 +192,57 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
...
+        'read_orientation': 'Direction of reads with respect to reference '
+                            'sequences in pre-trained sklearn classifier. '
+                            'same will cause reads to be classified unchanged'
+                            '; reverse-complement will cause reads to be '
+                            'reversed and complemented prior to '
+                            'classification. Default is to autodetect based '
+                            'on the confidence estimates for the first 100 '
+                            'reads.',
+        'reads_per_batch': 'Number of reads to process in each batch for '
+                           'sklearn classification. If 0, this parameter is '
```
```suggestion
                           'sklearn classification. If \'auto\', this parameter is '
```

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -242,3 +243,9 @@ def _compute_consensus_annotation(annotations, min_consensus,
         consensus_fraction_result = 1.0
 
     return annotation, consensus_fraction_result
+
+
+def _annotate_method(taxa, method):
```
thermokarst: I wish we had some way to ship "hidden" actions on plugins, this utility would be a great candidate for that (without it we break provenance). Add that one to the TODO list...

q2_feature_classifier/_vsearch.py
```diff
@@ -25,53 +30,159 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
...
+    if top_hits_only:
+        cmd.append('--top_hits_only')
+    cmd.append('--blast6out')
     consensus = _consensus_assignments(
         cmd, reference_taxonomy, min_consensus=min_consensus,
         unassignable_label=unassignable_label)
     return consensus
 
 
+def classify_hybrid_vsearch_sklearn(ctx,
```
thermokarst: A general note - should we start publishing the intermediate `FeatureData[Sequence]` & `FeatureData[Taxonomy]` created as part of our classifier training? Those Artifacts were helpful for me to have when testing this new Pipeline.

q2_feature_classifier/_vsearch.py
```diff
@@ -81,3 +192,57 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
...
+
+
+plugin.pipelines.register_function(
+    function=classify_hybrid_vsearch_sklearn,
+    inputs={**inputs, 'classifier': TaxonomicClassifier},
+    parameters={**parameters,
+                'reads_per_batch': Int % Range(0, None),
+                'confidence': Float,
+                'read_orientation': Str % Choices(
+                    ['same', 'reverse-complement']),
```
thermokarst: Should there be an `'auto'` choice, too? I recognize that that doesn't quite match up with the parent Action's API, but maybe this could let us move in the right direction.

q2_feature_classifier/_vsearch.py
```diff
@@ -81,3 +192,57 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
...
     citations=[citations['rognes2016vsearch']]
 )
+
+
+plugin.pipelines.register_function(
+    function=classify_hybrid_vsearch_sklearn,
+    inputs={**inputs, 'classifier': TaxonomicClassifier},
+    parameters={**parameters,
+                'reads_per_batch': Int % Range(0, None),
+                'confidence': Float,
```
```suggestion
                'confidence': Float % Range(0, 1, inclusive_start=False, inclusive_end=True) | Str % Choices(['disable', 'calculate_without_application']),
```

How does that look to you? I have no clue about the sentinel names, so edit away.

q2_feature_classifier/_vsearch.py
```diff
@@ -81,3 +192,57 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
...
+                            'reads.',
+        'reads_per_batch': 'Number of reads to process in each batch for '
+                           'sklearn classification. If 0, this parameter is '
+                           'autoscaled to min( number of query sequences / '
+                           'threads, 20000).',
+        'prefilter': 'Toggle positive filter of query sequences on or off.'
+    },
+    output_descriptions=output_descriptions,
+    name='ALPHA Hybrid classifier: VSEARCH exact match + sklearn classifier',
+    description=('NOTE: ALPHA RELEASE. Please report bugs to the q2-forum!\n'
```
```suggestion
    description=('NOTE: THIS PIPELINE IS AN ALPHA RELEASE. Please report bugs to https://forum.qiime2.org!\n'
```

q2_feature_classifier/_vsearch.py
thermokarst: A second general note - I ran this on MP tutorial and had 100% of the classification performed by sklearn. Should I look at another tutorial data set, or would you expect the same there?

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -242,3 +243,9 @@ def _compute_consensus_annotation(annotations, min_consensus,
...
+    taxa = taxa.view(pd.DataFrame)
+    taxa['method'] = method
```
thermokarst:
```suggestion
    taxa['Method'] = method
```

image

I am being a bit opinionated here...

q2_feature_classifier/_vsearch.py
nbokulich: LGTM!

q2_feature_classifier/_vsearch.py
nbokulich: thx!

q2_feature_classifier/_consensus_assignment.py
nbokulich: yes! we've had this discussion before re: actions in q2-sample-classifier and q2-longitudinal that should be "hidden". I suppose `annotate_method` could be usefully exposed as a method, allowing users to reconstruct their own hybrid classifier pipelines...

thermokarst: > I suppose annotate_method could be usefully exposed as a method

Your call :call_me_hand:

q2_feature_classifier/_vsearch.py
nbokulich: intermediate files == ref sequences? Would not hurt to publish (though then we get back into the issue of database devs not wanting us to re-host their data). If the classifier training tutorial were on the 99% OTUs instead of the 88%, we would have intermediate files and classifier trained with the current sklearn release every release... but then it bogs down busywork. I am ranting now.

> I ran this on MP tutorial and had 100% of the classification performed by sklearn. Should I look at another tutorial data set, or would you expect the same there?
Maybe try out on another dataset? On some of my own data I was finding majority VSEARCH classification... it may just depend on the dataset.

q2_feature_classifier/_vsearch.py
nbokulich: what would 'auto' do? I agree it would be a move in the right direction but right now the `classify_sklearn` method has no such option.

q2_feature_classifier/_consensus_assignment.py
nbokulich: let's leave it hidden for now

q2_feature_classifier/_vsearch.py
nbokulich: again, trying to mirror the underlying params in this pipeline — and the underlying params need an update, so let me push up another commit with those changes.

q2_feature_classifier/_vsearch.py
nbokulich: 👍

q2_feature_classifier/_vsearch.py
nbokulich: 👍

q2_feature_classifier/_consensus_assignment.py
nbokulich: 🍅 

lol fine

nbokulich:
thanks @thermokarst ! Very sneaky of you to use this as an opportunity to have me clean up @BenKaehler 's munted parameter definitions 😉 

I have mostly made all changes you recommend — I think you will be pleased with the changes.

I am praying tests pass.

q2_feature_classifier/classifier.py
```diff
@@ -226,9 +226,35 @@ def classify_sklearn(reads: DNAFASTAFormat, classifier: Pipeline,
     'reads_per_batch': Int % Range(0, None),
     'n_jobs': Int,
     'pre_dispatch': Str,
-    'confidence': Float,
-    'read_orientation': Str % Choices(['same', 'reverse-complement'])}
-
+    'confidence': Float % Range(
+        0, 1, inclusive_start=True, inclusive_end=True) | Str % Choices(
```
thermokarst: `0` still feels a bit magical to me, but maybe with some additional explanation I can be steered in the right direction? Unless we have a reason for it, my vote is to move that to a `Str Choices` option.

q2_feature_classifier/_vsearch.py
```diff
@@ -25,53 +31,161 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
...
+
+input_descriptions = {'query': 'Sequences to classify taxonomically.',
+                      'reference_reads': 'reference sequences.',
+                      'reference_taxonomy': 'reference taxonomy labels.'}
+
+parameter_descriptions = {
+    'strand': 'Align against reference sequences in forward ("plus") '
+              'or both directions ("both").',
+    'maxaccepts': 'Maximum number of hits to keep for each query. Set to '
+                  '"all" to keep all hits > perc_identity similarity. Must '
```
thermokarst: Just a general note, I think that communicating range constraints is generally a job for the interface. No worries on including it here (it can't hurt, and might help teach people how to read the type annotations!), but just wanted to toss that out there for future work.

q2_feature_classifier/classifier.py
```diff
@@ -226,9 +226,35 @@ def classify_sklearn(reads: DNAFASTAFormat, classifier: Pipeline,
...
+                        'respect to reference sequences. same will cause '
+                        'reads to be classified unchanged; reverse-'
+                        'complement will cause reads to be reversed '
+                        'and complemented prior to classification. '
+                        '"auto" will autodetect orientation based on the '
+                        'confidence estimates for the first 100 reads.',
+    'reads_per_batch': 'Number of reads to process in each batch. If 0, '
+                       'this parameter is autoscaled to '
+                       'min( number of query sequences / n_jobs, 20000).',
+    'n_jobs': 'The maximum number of concurrently worker processes. If -1 '
```
thermokarst: Does this description need to be updated for `'auto'`?

q2_feature_classifier/classifier.py
```diff
@@ -226,9 +226,35 @@ def classify_sklearn(reads: DNAFASTAFormat, classifier: Pipeline,
...
+                  'confidence but not apply it to limit the '
+                  'taxonomic depth of the assignments.',
+    'read_orientation': 'Direction of reads with '
+                        'respect to reference sequences. same will cause '
+                        'reads to be classified unchanged; reverse-'
+                        'complement will cause reads to be reversed '
+                        'and complemented prior to classification. '
+                        '"auto" will autodetect orientation based on the '
+                        'confidence estimates for the first 100 reads.',
+    'reads_per_batch': 'Number of reads to process in each batch. If 0, '
```
thermokarst: Does this description need to be updated for `'auto'`?

q2_feature_classifier/_vsearch.py
```diff
@@ -25,53 +31,161 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
...
+output_descriptions = {
+    'classification': 'The resulting taxonomy classifications.'}
+
+parameters = {'maxaccepts': Int % Range(1, None) | Str % Choices(['all']),
+              'perc_identity': Float % Range(0.0, 1.0, inclusive_end=True),
+              'query_cov': Float % Range(0.0, 1.0, inclusive_end=True),
+              'strand': Str % Choices(['both', 'plus']),
+              'min_consensus': Float % Range(0.5, 1.0, inclusive_end=True,
+                                             inclusive_start=False),
+              'threads': Int}
```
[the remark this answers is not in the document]
nbokulich: 👍

q2_feature_classifier/classifier.py
nbokulich: `0` is to match the sklearn API. If we introduce a `Calculate_but_do_not_use` option, it would just necessitate a couple extra lines of code to make `confidence=0`.

thermokarst: Matching sklearn API sounds like a great reason to leave as-is - thanks for explaining!

q2_feature_classifier/_vsearch.py
nbokulich: Yep, this is an old parameter description (Nick's first plugin method 👶) — I just moved these parameter descriptions out of the `classify_consensus_vsearch` registration without reading them!

I am happy to change now that we are shining some lights in the ol' gear box here.

q2_feature_classifier/classifier.py
nbokulich: thanks for catching

q2_feature_classifier/classifier.py
nbokulich: 👍

Issue #135, opened 2019-06-11
`extract-reads`: add n-jobs parameter


ebolyen:
**Improvement Description**
It looks like it may be possible to meaningfully parallelize `extract-reads` based on [this forum post](https://forum.qiime2.org/t/specifying-how-many-threads-to-run-on-with-qiime2-feature-classifier-extract-reads/10111/5).

We would need this to be done via some kind of multi-processing workflow (hence jobs not threads), but joblib might make this easy-ish.

**Questions**
1. At some point we will definitely become IO-bound, what is that point? Should there be a maximum set for the number of jobs, since even an NVMe SSD will eventually tap out.

**Comments**
1. It may be worth verifying CPU vs disk IO before starting this in earnest.

**References**
1. https://forum.qiime2.org/t/specifying-how-many-threads-to-run-on-with-qiime2-feature-classifier-extract-reads/10111/5

BenKaehler:
This has been on my todo list for ages.

The process only has IO at the start and end and it is negligible.

I will parallelise it consistent with read classification (using joblib).

Oddant1:
I'm interested in taking a stab at this issue

Issue #134, opened 2019-05-17
joblib now a dependency in scikit-learn


ChrisKeefe:
**Bug Description**
Scikit-learn recently [spun off `joblib` as a dependency](https://scikit-learn.org/stable/whats_new.html#miscellaneous), removing it from the scikit-learn codebase. 

When QIIME 2 v2019.4 is updated with `scikit-learn v21.1`,  `q2-feature classifier` throws a deprecation warning when running `fit-classifier-naive-bayes`:
```
/scratch/crk239/conda-envs/q2-19.4/lib/python3.6/site-packages/sklearn/externals/joblib/__init__.py:15: DeprecationWarning: $
 warnings.warn(msg, category=DeprecationWarning)
```

This has not been extensively tested, but likely impacts other sklearn-based Actions as well.
q2-feature-classifier should be refactored as needed to handle the change of `joblib` from vendored to dependency software. 

**Expected behavior**
`fit-classifier-naive-bayes` (and other Actions) should complete without deprecation warnings

**Computation Environment**
- OS: CentOS 6.10
- slurm 18.08.7
- QIIME 2 Release 2019.4

**References**
1. https://scikit-learn.org/stable/whats_new.html#miscellaneous

thermokarst:
cc @nbokulich && @BenKaehler

thermokarst:
Let's address this in 2019.7.

PR #136, opened 2019-06-14
IMP: Parallelize extract_reads

Oddant1:
closes: #135


ebolyen:
Sorry lost track of this, it's on the board now. I'll take a first pass today, and @BenKaehler would you mind doing a second pass?

ebolyen:
Update, @thermokarst will take a first pass!

q2_feature_classifier/_cutter.py
```diff
@@ -178,7 +200,9 @@ def extract_reads(sequences: DNAIterator,  f_primer: str, r_primer: str,
                 'r_primer': Str,
                 'identity': Float,
                 'min_length': Int % Range(0, None),
-                'max_length': Int % Range(0, None)},
+                'max_length': Int % Range(0, None),
+                'n_jobs': Int % Range(1, None),
+                'batch_size': Int % Range(0, None)},
```
thermokarst: How about this, instead:

```suggestion
               Int % Range(1, None) | Str % Choices(['auto']),
```


cc @ebolyen, does this look right to you? I can't seem to get this to work locally (`AttributeError: <class 'qiime2.core.type.grammar.UnionExp'> object has no attribute 'predicate'`)

q2_feature_classifier/_cutter.py
```diff
@@ -203,5 +227,11 @@ def extract_reads(sequences: DNAIterator,  f_primer: str, r_primer: str,
                                           'before trimming and truncation, '
                                           'so plan accordingly. Set to zero '
                                           '(default) to disable max length '
-                                          'filtering.'}
+                                          'filtering.',
+                            'n_jobs': 'Number of seperate processes to run. '
+                                      'Default is one.',
+                            'batch_size': 'Number of sequences per batch '
```
thermokarst:
```
Number of sequences to process in a batch.
The `auto` option is calculated from the
number of sequences and number of jobs
specified.
```

q2_feature_classifier/_cutter.py
```diff
@@ -203,5 +227,11 @@ def extract_reads(sequences: DNAIterator,  f_primer: str, r_primer: str,
                                           'before trimming and truncation, '
                                           'so plan accordingly. Set to zero '
                                           '(default) to disable max length '
-                                          'filtering.'}
+                                          'filtering.',
+                            'n_jobs': 'Number of seperate processes to run. '
+                                      'Default is one.',
```
thermokarst: No need to mention the default value here --- interfaces should indicate the default if it exists.

q2_feature_classifier/_cutter.py
thermokarst: Just to follow up, this only raises the `AttributeError` in a plugin registration, it is fine if I construct manually and interrogate:

```python
In [5]: x = Int % Range(1, None) | Str % Choices(['auto'])

In [6]: type(x)
Out[6]: qiime2.core.type.grammar.UnionExp

In [7]: x
Out[7]: Int % Range(1, None) | Str % Choices('auto')

In [8]: x.members
Out[8]: (Int % Range(1, None), Str % Choices('auto'))

In [9]: 'auto' in x
Out[9]: True

In [10]: 'fauto' in x
Out[10]: False

In [11]: 2 in x
Out[11]: True
```

ebolyen: Oh dear... the plugin API might be touching things it ideally shouldn't...

thermokarst: You think so? I am still willing to chalk it up to dev error. I could've _sworn_ we tested this case out when working on types v4....

thermokarst:
BTW, @Oddant1 - what happened with the permission change on `dna-sequences.fasta`?

Oddant1:
@thermokarst I have absolutely no idea, but I'll change it back

Oddant1:
https://github.com/qiime2/q2-feature-classifier/pull/136/commits/01126068d13873f969446416b662dae7873c8598 perhaps worth noting, all this commit did was revert the anomalous (I have no idea why it happened) permission change to that data file. This inexplicably failed in travis, but the next commit which added additional changes on top of that permission reversion passed.

thermokarst:
Hey @Oddant1 --- I am going to fix the blocking issue in https://github.com/qiime2/q2cli/issues/221 before we proceed with the remaining review here. Stay tuned! :tv:

q2_feature_classifier/_cutter.py
thermokarst: Issue: https://github.com/qiime2/q2cli/issues/221

Oddant1:
@thermokarst well that looks like fun :laughing:

q2_feature_classifier/_cutter.py
```diff
@@ -6,11 +6,16 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
-from itertools import chain
+import skbio
+import os
+from sklearn.externals.joblib import Parallel, delayed, effective_n_jobs
```
nbokulich: see #134 . change to:
```from joblib import Parallel, delayed, effective_n_jobs```

Comment (Oddant1) [reply q2_feature_classifier/_cutter.py]: @nbokulich I made it `from sklearn.externals.joblib` not `from joblib` because `from joblib` wasn't working. After reading your comment, I tried changing it to `from joblib` again, and it did work. WHAT KIND OF BLACK MAGIC ARE YOU CONDUCTING

**Comment by thermokarst** [comment]
Hey @Oddant1, we fixed the blocking issue in q2cli, this PR is back in your court now.

File: q2_feature_classifier/_cutter.py (hunk 1 above)
Comment (Oddant1) [reply q2_feature_classifier/_cutter.py]: @thermokarst How exactly is this feature meant to work? What type of object is `Choices` supposed to be?

File: q2_feature_classifier/_cutter.py [hunk 5]
```diff
@@ -98,75 +103,92 @@ def _approx_match(seq, f_primer, r_primer, identity):
...
+        return
+    return amp
+
+
+def extract_reads(sequences: DNASequencesDirectoryFormat,  f_primer: str,
+                  r_primer: str, trunc_len: int = 0, trim_left: int = 0,
                   identity: float = 0.8, min_length: int = 50,
-                  max_length: int = 0) -> DNAIterator:
+                  max_length: int = 0, n_jobs: int = 1,
+                  batch_size: int = 'auto') -> DNAFASTAFormat:
```
thermokarst: The type annotation doesn't matter here. Personally, I prefer to set this type to whatever the default value's type is, but, its six of one, half-dozen of the other.

PR #140, opened 2019-06-20
DEP: Unpin sklearn

thermokarst:
How about this incantation?


ebolyen:
I have to say, githubs commit graph here is super confusing.

thermokarst:
Yeah, weird that the reference node looks identical to an actual changeset node.

PR #145, opened 2019-07-30
make hybrid classifier test result more determinate

nbokulich:
Added randseed, sample_size, and smoothed out param descriptions

we were getting test failure 1/10 times... the pre-filter appears to have been the problem.


q2_feature_classifier/_vsearch.py
```diff
@@ -156,6 +159,8 @@ def classify_hybrid_vsearch_sklearn(ctx,
 
 outputs = [('classification', FeatureData[Taxonomy])]
 
+ignore_prefilter = ' This parameter is ignored if prefilter==False.'
```
```suggestion
ignore_prefilter = ' This parameter is ignored if `prefilter` is disabled.'
```

Issue #131, opened 2019-04-01
`classify-sklearn`: MemoryError should be more user-friendly


nbokulich:
**Improvement Description**
Report some troubleshooting steps in the error message

**Current Behavior**
the raw `MemoryError` message is reported

**Proposed Behavior**
Catch memory error and give a better message. Mention:
1. `reads-per-batch` parameter
2. issues with parallel jobs
3. try a smaller database
4. get a more powerful machine or allocate more resources

David-Rod:
Hey @nbokulich, I know that you are currently working on the new release, but I was hoping to get some more context for this issue. Would you be able to provide the steps and files that are needed to replicate the `MemoryError` for this issue? I have been trying to replicate on my own, but I have been unsuccessful so far. Thanks.

nbokulich:
hey @David-Rod to be honest I don't have a good test example because I have not run into this issue before (it is usually encountered if you have a weak system or unusually large data).

I recommend grabbing the full-length SILVA classifier and classifying any 16S sequences. A great way to eat up more memory is to run with `n_jobs > 1`

David-Rod:
@nbokulich  So I used the full-length SILVA classifier, and I received the following:

![image](https://user-images.githubusercontent.com/30784905/62166993-013d0600-b2d7-11e9-8319-e20004c524f5.png)

Does this look similar to the error you might have received? I have received no messages in my terminal as of now. My machine is just spinning.

nbokulich:
yikes! no I have not encountered that before. It is surprising the
MemoryError has not been raised.

thermokarst:
Hi @David-Rod --- that looks like a _disk space_ error, not a memory (RAM) error. One possibility for testing memory errors is to adjust the ulimit settings on your development machine to artificially "lower" the available memory.

nbokulich:
is this issue resolved? I see there was some discussion over whether @David-Rod 's changes covered all memory errors, but it looks like @ebolyen 's changes may have fixed this?

PR #146, opened 2019-08-05
ENH: Initial commit to improve the memory error

David-Rod:
This initial commit is provides a more clear `MemoryError` message. This PR is meant to resolve issue #131. 

OSX seems to bypass RAM, once memory has been exceeded, and will start to use storage space. I attempted to artificially lower the memory available for this process, but OSX disregards the use of `ulimit` commands. I could not recreate the issue on OSX because of this.

I was able to recreate the memory issue on Linux Ubuntu. There was no need to artificially limit memory using Linux. The terminal produces the `MemoryError` after allowing this process to consume available memory.

### DIFF
--- a/q2_feature_classifier/_skl.py
+++ b/q2_feature_classifier/_skl.py
@@ -39,10 +39,12 @@ def _extract_reads(reads):
 
 def predict(reads, pipeline, separator=';', chunk_size=262144, n_jobs=1,
             pre_dispatch='2*n_jobs', confidence='disable'):
-    return (m for c in Parallel(n_jobs=n_jobs, batch_size=1,
-                                pre_dispatch=pre_dispatch)
-            (delayed(_predict_chunk)(pipeline, separator, confidence, chunk)
-             for chunk in _chunks(reads, chunk_size)) for m in c)
+    jobs = (
+        delayed(_predict_chunk)(pipeline, separator, confidence, chunk)
+        for chunk in _chunks(reads, chunk_size))
+    workers = Parallel(n_jobs=n_jobs, batch_size=1, pre_dispatch=pre_dispatch)
+    for calculated in workers(jobs):
+        yield from calculated
 
 
 def _predict_chunk(pipeline, separator, confidence, chunk):
--- a/q2_feature_classifier/classifier.py
+++ b/q2_feature_classifier/classifier.py
@@ -203,34 +203,33 @@ def classify_sklearn(reads: DNAFASTAFormat, classifier: Pipeline,
                      pre_dispatch: str = '2*n_jobs', confidence: float = 0.7,
                      read_orientation: str = 'auto'
                      ) -> pd.DataFrame:
-    # autotune reads per batch
-    if reads_per_batch == 0:
-        reads_per_batch = _autotune_reads_per_batch(reads, n_jobs)
-
-    # transform reads to DNAIterator
-    reads = DNAIterator(
-        skbio.read(str(reads), format='fasta', constructor=skbio.DNA))
-
-    reads = _autodetect_orientation(
-        reads, classifier, read_orientation=read_orientation)
-    predictions = predict(reads, classifier, chunk_size=reads_per_batch,
-                          n_jobs=n_jobs, pre_dispatch=pre_dispatch,
-                          confidence=confidence)
     try:
+        # autotune reads per batch
+        if reads_per_batch == 0:
+            reads_per_batch = _autotune_reads_per_batch(reads, n_jobs)
+
+        # transform reads to DNAIterator
+        reads = DNAIterator(
+            skbio.read(str(reads), format='fasta', constructor=skbio.DNA))
+
+        reads = _autodetect_orientation(
+            reads, classifier, read_orientation=read_orientation)
+        predictions = predict(reads, classifier, chunk_size=reads_per_batch,
+                              n_jobs=n_jobs, pre_dispatch=pre_dispatch,
+                              confidence=confidence)
         seq_ids, taxonomy, confidence = list(zip(*predictions))
+
+        result = pd.DataFrame({'Taxon': taxonomy, 'Confidence': confidence},
+                              index=seq_ids, columns=['Taxon', 'Confidence'])
+        result.index.name = 'Feature ID'
+        return result
     except MemoryError:
         raise MemoryError("The operation has run out of available memory. "
                           "To correct this error:\n"
                           "1. Reduce the reads per batch\n"
                           "2. Reduce number of n_jobs being performed\n"
-                          "3. Try using a smaller data set\n"
-                          "4. Use a more powerful machine or allocate "
+                          "3. Use a more powerful machine or allocate "
                           "more resources ")
-
-    result = pd.DataFrame({'Taxon': taxonomy, 'Confidence': confidence},
-                          index=seq_ids, columns=['Taxon', 'Confidence'])
-    result.index.name = 'Feature ID'
-    return result
 
 
 _classify_parameters = {
--- a/q2_feature_classifier/classifier.py
+++ b/q2_feature_classifier/classifier.py
@@ -216,7 +216,17 @@ def classify_sklearn(reads: DNAFASTAFormat, classifier: Pipeline,
     predictions = predict(reads, classifier, chunk_size=reads_per_batch,
                           n_jobs=n_jobs, pre_dispatch=pre_dispatch,
                           confidence=confidence)
-    seq_ids, taxonomy, confidence = list(zip(*predictions))
+    try:
+        seq_ids, taxonomy, confidence = list(zip(*predictions))
+    except MemoryError:
+        raise MemoryError("The operation has run out of available memory. "
+                          "To correct this error:\n"
+                          "1. Reduce the reads per batch\n"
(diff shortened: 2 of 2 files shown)


q2_feature_classifier/_taxonomic_classifier.py
```diff
@@ -69,7 +69,17 @@ def _1(dirfmt: TaxonomicClassiferTemporaryPickleDirFmt) -> Pipeline:
         tmpdir = model.DirectoryFormat()
         dirname = str(tmpdir)
         tar.extractall(dirname)
-        pipeline = joblib.load(os.path.join(dirname, 'sklearn_pipeline.pkl'))
+        try:
+            pipeline = joblib.load(os.path.join(dirname,
+                                                'sklearn_pipeline.pkl'))
+        except MemoryError:
```
thermokarst: Let's add a test for this, by [mocking](https://docs.python.org/3.6/library/unittest.mock.html) joblib.load to raise this memory error. Please let one of us know if you need a hand with this!

q2_feature_classifier/_taxonomic_classifier.py
```diff
@@ -69,7 +69,17 @@ def _1(dirfmt: TaxonomicClassiferTemporaryPickleDirFmt) -> Pipeline:
...
+            raise MemoryError("There is a resource issue regarding your "
```
thermokarst: How about we replace the first sentence with something like what is in the [python stdlib docs](https://docs.python.org/3/library/exceptions.html#MemoryError):

> The operation has run out of available memory. To correct this error:...

q2_feature_classifier/_taxonomic_classifier.py
David-Rod: I was able to work on this with @Oddant1. If you like the changes, I would like to add him as a co-author.

q2_feature_classifier/_taxonomic_classifier.py
David-Rod: I changed the error message as requested.

q2_feature_classifier/_taxonomic_classifier.py
```diff
@@ -69,7 +69,17 @@ def _1(dirfmt: TaxonomicClassiferTemporaryPickleDirFmt) -> Pipeline:
...
+            pipeline = joblib.load(os.path.join(dirname,
+                                                'sklearn_pipeline.pkl'))
+        except MemoryError:
+            raise MemoryError("The operation has run out of available memory. "
+                              "To correct this error:\n"
+                              "1. Reduce the reads per batch\n"
+                              "2. Reduce number of n_jobs being performed\n"
+                              "3. Try using a smaller data set\n"
+                              "4. Use a more powerful machine or allocate "
+                              "more resources ")
```
thermokarst:
```suggestion
                              "more resources")
```

q2_feature_classifier/_taxonomic_classifier.py
thermokarst: Is this where the action generally fails at? Looking around on the forum, it seems like more often than not, the loading succeeds, but the action fails later when running the meat of the command. Its possible your VBox testing is triggering this failure point, which is a good catch, but, I suspect it isn't the typical failure point.

q2_feature_classifier/_taxonomic_classifier.py
David-Rod: Hello @thermokarst,

Regarding your comment on the `MemoryError` placement https://github.com/qiime2/q2-feature-classifier/pull/146#discussion_r314782289,

I have been searching through the forum to determine where the `MemoryError` might be occurring. If the load step is not the issue, and if the user is not specifying `n-jobs` (leads me to believe that the number of jobs is not necessarily the direct cause of the error), does that mean that the problem might be in `classifier.py` rather than the transformer?

thermokarst: Looks like this message might've gotten posted in the wrong thread, I will follow up in the original thread.

q2_feature_classifier/_taxonomic_classifier.py
thermokarst: I searched for the error on the forum, and one of the first hits was: https://forum.qiime2.org/t/memory-error-qiime-feature-classifier-classify-sklearn/9066

If you look at the error traceback, the first "qiime 2" code in the call stack is:

```bash
File "/home/patricia/miniconda2/envs/qiime2-2019.1/lib/python3.6/site-packages/q2_feature_classifier/
skl.py ", line 45, in predict
```

That means that this is the "last" bit of the plugin's code to be called before the error crops up. This seems like a good place to start.

q2_feature_classifier/_taxonomic_classifier.py
David-Rod: For some reason, I couldn't respond to that specific thread, so I decided to submit my question ( for the time being) in this one. Thanks!

thermokarst: image

My PR view shows a reply box when I scroll _up_, which is weird. YMMV.

q2_feature_classifier/_skl.py
```diff
@@ -39,10 +39,12 @@ def _extract_reads(reads):
...
-    return (m for c in Parallel(n_jobs=n_jobs, batch_size=1,
-                                pre_dispatch=pre_dispatch)
-            (delayed(_predict_chunk)(pipeline, separator, confidence, chunk)
-             for chunk in _chunks(reads, chunk_size)) for m in c)
+    jobs = (
+        delayed(_predict_chunk)(pipeline, separator, confidence, chunk)
+        for chunk in _chunks(reads, chunk_size))
+    workers = Parallel(n_jobs=n_jobs, batch_size=1, pre_dispatch=pre_dispatch)
+    for calculated in workers(jobs):
+        yield from calculated
```
Oddant1: Yay, it's readable now!

q2_feature_classifier/classifier.py
```diff
@@ -216,7 +216,17 @@ def classify_sklearn(reads: DNAFASTAFormat, classifier: Pipeline,
...
+        seq_ids, taxonomy, confidence = list(zip(*predictions))
+    except MemoryError:
+        raise MemoryError("The operation has run out of available memory. "
+                          "To correct this error:\n"
+                          "1. Reduce the reads per batch\n"
+                          "2. Reduce number of n_jobs being performed\n"
+                          "3. Try using a smaller data set\n"
+                          "4. Use a more powerful machine or allocate "
+                          "more resources ")
+
```
Oddant1: @thermokarst I attempted to raise this error on my machine, and was unable to. @David-Rod says he still got the old memory error on the load step when he tried to get this error, but for me the command runs fine regardless of how many jobs I allocate (I tried some sane numbers like 4-12 and some less sane numbers like 40-100).

Oddant1: Addendum. I was able to raise _A_ `MemoryError` after leaving the process running for about half an hour. This is the traceback.
```Traceback (most recent call last):
  File "/home/anthony/Desktop/src/Qiime2/q2cli/q2cli/commands.py", line 327, in __call__
    results = action(**arguments)
  File "</home/anthony/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/decorator.py:decorator-gen-467>", line 2, in classify_sklearn
  File "/home/anthony/Desktop/src/Qiime2/qiime2/qiime2/sdk/action.py", line 229, in bound_callable
    spec.view_type, recorder)
  File "/home/anthony/Desktop/src/Qiime2/qiime2/qiime2/sdk/result.py", line 289, in _view
    result = transformation(self._archiver.data_dir)
  File "/home/anthony/Desktop/src/Qiime2/qiime2/qiime2/core/transform.py", line 70, in transformation
    new_view = transformer(view)
  File "/home/anthony/Desktop/src/Qiime2/q2-feature-classifier/q2_feature_classifier/_taxonomic_classifier.py", line 72, in _1
    pipeline = joblib.load(os.path.join(dirname, 'sklearn_pipeline.pkl'))
  File "/home/anthony/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/joblib/numpy_pickle.py", line 598, in load
    obj = _unpickle(fobj, filename, mmap_mode)
  File "/home/anthony/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/joblib/numpy_pickle.py", line 526, in _unpickle
    obj = unpickler.load()
  File "/home/anthony/miniconda3/envs/qiime2-dev/lib/python3.6/pickle.py", line 1050, in load
    dispatch[key[0]](self)
  File "/home/anthony/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/joblib/numpy_pickle.py", line 352, in load_build
    self.stack.append(array_wrapper.read(self))
  File "/home/anthony/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/joblib/numpy_pickle.py", line 195, in read
    array = self.read_array(unpickler)
  File "/home/anthony/miniconda3/envs/qiime2-dev/lib/python3.6/site-packages/joblib/numpy_pickle.py", line 141, in read_array
    array = unpickler.np.empty(count, dtype=self.dtype)
MemoryError
```
This is not the `MemoryError` observed by forum users, this is the same `MemoryError` @David-Rod was initially getting. I will experiment further to see if I can replicate forum behavior.

PR #148, opened 2019-08-07
Check for min length > trunc length before executing

### LINKED ISSUE #119: `extract-reads`: validate trimming parameters prior to extraction
**Improvement Description**
If `min-length` > `trunc-len`, all reads will be filtered out. This should be caught at the outset, before any read extraction occurs.

**Current Behavior**
Extraction will run (slow), then trimming, then `min-length` filtering.

**Proposed Behavior**
Either raise an error at the outset if `min-length` > `trunc-len`, otherwise reconsider order of trimming and filtering parameters (should we extract -> filter -> trim instead?).

Oddant1:
Closes #119. A ValueError is now raised immediately if `min_length` is greater than `trunc_len` and `trunc_len` is greater than 0.

### DIFF
--- a/q2_feature_classifier/_cutter.py
+++ b/q2_feature_classifier/_cutter.py
@@ -168,10 +168,10 @@ def extract_reads(sequences: DNASequencesDirectoryFormat, f_primer: str,
     q2_types.DNAFASTAFormat
         containing the reads
     """
-    if min_length > trunc_len and trunc_len > 0:
+    if min_length > trunc_len - trim_left and trunc_len > 0:
         raise ValueError('Your minimum length is larger than your truncation '
-                         'length, this will result in all of your sequences '
-                         'being trimmed automatically.')
+                         'length minus your trim, this will result in all of '
+                         'your sequences being trimmed automatically.')
     n_jobs = effective_n_jobs(n_jobs)
     if batch_size == 'auto':
         batch_size = _autotune_reads_per_batch(
--- a/q2_feature_classifier/_cutter.py
+++ b/q2_feature_classifier/_cutter.py
@@ -168,6 +168,10 @@ def extract_reads(sequences: DNASequencesDirectoryFormat, f_primer: str,
     q2_types.DNAFASTAFormat
         containing the reads
     """
+    if min_length > trunc_len and trunc_len > 0:
+        raise ValueError('Your minimum length is larger than your truncation '
+                         'length, this will result in all of your sequences '
+                         'being trimmed automatically.')
     n_jobs = effective_n_jobs(n_jobs)
     if batch_size == 'auto':
         batch_size = _autotune_reads_per_batch(
--- a/q2_feature_classifier/_cutter.py
+++ b/q2_feature_classifier/_cutter.py
@@ -169,11 +169,10 @@ def extract_reads(sequences: DNASequencesDirectoryFormat, f_primer: str,
         containing the reads
     """
     if min_length > trunc_len - trim_left and trunc_len > 0:
-        raise ValueError('Your minimum length is larger than your truncation '
-                         'length minus your trim. This will result in all of '
-                         'your sequences being removed from your dataset. To '
-                         'procees, you must set a minimum length less than '
-                         'your truncation length minus your trim.')
+        raise ValueError('The minimum length setting is greater than the '
+                         'length of the truncated sequences. This will cause '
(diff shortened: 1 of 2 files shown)


q2_feature_classifier/_cutter.py
```diff
@@ -168,6 +168,10 @@ def extract_reads(sequences: DNASequencesDirectoryFormat, f_primer: str,
     q2_types.DNAFASTAFormat
         containing the reads
     """
+    if min_length > trunc_len and trunc_len > 0:
+        raise ValueError('Your minimum length is larger than your truncation '
+                         'length, this will result in all of your sequences '
```
nbokulich: "length, this" --> "length. This"

q2_feature_classifier/_cutter.py
```diff
@@ -168,6 +168,10 @@ def extract_reads(sequences: DNASequencesDirectoryFormat, f_primer: str,
...
+                         'being trimmed automatically.')
```
nbokulich: "trimmed automatically." --> "removed from your dataset. To proceed, you must set a minimum length less than your truncation length."

q2_feature_classifier/_cutter.py
```diff
@@ -168,6 +168,10 @@ def extract_reads(sequences: DNASequencesDirectoryFormat, f_primer: str,
     q2_types.DNAFASTAFormat
         containing the reads
     """
+    if min_length > trunc_len and trunc_len > 0:
+        raise ValueError('Your minimum length is larger than your truncation '
```
nbokulich: "larger than" --> "greater than or equal to"

q2_feature_classifier/tests/test_cutter.py
```diff
@@ -98,7 +98,7 @@ def test_extract_reads_fail_max_length(self):
                 max_length=1)
 
     def test_extract_reads_fail_trim_entire_read(self):
```
nbokulich: good catch that this test will trigger this error. However, it looks like the purpose of this test was to error out if all read lengths == 0.

Please split these into two tests. `test_extract_reads_fail_trim_entire_read` should use `min_length=0` so that we are testing for the original error message. Please add a new test for `min_length>trunc_len`

q2_feature_classifier/_cutter.py
Oddant1: I misinterpreted this initially, and changed the check to `min_len >= trunc_len` instead of changing the comment, but now that I think about it, will it be a problem if `min_len == trunc_len`? Shouldn't it only be an issue if the minimum length is greater?

Oddant1: Also, just to be clear, on that last commit the check has been changed to `min_len >= trunc_len` and the message is unchanged in accordance with my initial misinterpretation of your comment. Easy enough to fix though obviously.

nbokulich: good call @Oddant1 I think I was misinterpreting this before.

But the plot thickens... it occurs to me that this validation should also include trim_left, since that will impact the length. Should this instead be `if min_length > trunc_len - trim_left`? And also change the error message accordingly.

q2_feature_classifier/_cutter.py
```diff
@@ -170,8 +170,10 @@ def extract_reads(sequences: DNASequencesDirectoryFormat, f_primer: str,
     """
     if min_length > trunc_len - trim_left and trunc_len > 0:
         raise ValueError('Your minimum length is larger than your truncation '
-                         'length minus your trim, this will result in all of '
-                         'your sequences being trimmed automatically.')
+                         'length minus your trim. This will result in all of '
+                         'your sequences being removed from your dataset. To '
+                         'procees, you must set a minimum length less than '
+                         'your truncation length minus your trim.')
```
nbokulich: the wording is a bit awkward on this error message now. How about this:
"The minimum length setting is greater than the length of the truncated sequences. This will cause all sequences to be removed from the dataset. To proceed, set a min_length ≤ trunc_len - trim_left."

Oddant1:
I'm going to add one more test where I pass a `vmin` value greater than `vmax` just for good measure

nbokulich:
> I'm going to add one more test where I pass a vmin value greater than vmax just for good measure

did this comment go in the correct PR?

Or did you mean `min_length` and `trunc_len`?

Oddant1:
@nbokulich no I accidentally put that in the wrong PR :laughing:

PR #153, opened 2019-10-21
added maxhits maxrejects weak_id and other options to classify-consensus-vsearch

### LINKED ISSUE #151: classify-consensus-vsearch: expose maxhits option (and other options)
**Improvement Description**
Expose vsearch maxhits option, which explicitly controls the number of hits that are reported. 

While we're at it, expose some more options to make behavior more flexible: `maxrejects` (default 0), `output_no_hits` (default `True`), `weak_id` (default None), anything else?

**Current Behavior**
maxaccepts (currently exposed) implicitly controls the number of hits reported, but will return > 1 hit if `strand=both` and two different sequences are hit. See: https://github.com/torognes/vsearch/issues/388#issuecomment-539009338

**Proposed Behavior**
maxhits would make it very clear how many hits are reported, e.g., supporting arbitrary top-hit searches.

nbokulich:
Fixes #151 

Added maxhits, maxrejects, output_no_hits, and weak_id options to classify-consensus-vsearch and the hybrid classifier, and a simple test to make sure they don't break anything.


colinbrislawn:
This might be a separate discussion, but now is a great time to update the vsearch defaults.

Suggested by @torognes on [the forums](https://groups.google.com/d/msg/vsearch-forum/xwlzDbL91Cc/q3I01DTAAAAJ):
> Using something like --maxaccepts 100 --maxrejects 100 --maxhits 1 --id 0.77

[This pipeline](https://github.com/pnnl/hundo/blob/master/hundo/Snakefile#L742-L744) uses similar defaults:
>vsearch --maxaccepts 25 --maxrejects 100 --id 0.6

Let me know if you think these are 'sensible defaults.'

thermokarst:
> Suggested by @torognes on [the forums](https://groups.google.com/d/msg/vsearch-forum/xwlzDbL91Cc/q3I01DTAAAAJ):
> 
> > Using something like --maxaccepts 100 --maxrejects 100 --maxhits 1 --id 0.77

As a general rule, when we expose 3rd party tools (like vsearch), we keep the same defaults that the tools have set for themselves. It sounds to me like the values you shared above _aren't_ vsearch's default values. We would need to come up with a solid justification as to _why_ q2-feature-classifier is changing tool's defaults.

nbokulich:
> We would need to come up with a solid justification as to why q2-feature-classifier is changing tool's defaults.

q2-feature-classifier's defaults are all based on our published benchmarking results. Standalone VSEARCH was not designed or benchmarked for taxonomy classification per se, so their defaults needn't necessarily be used for a taxonomy classification tool that uses VSEARCH for alignment.

The new parameters added in this PR were not benchmarked, but the current defaults are based either on what we previously tested or else on vsearch defaults (as in the case of maxrejects)

colinbrislawn:
For future ref, here's the current vsearch defaults:
```text
--maxaccepts 1
--maxrejects 32
--maxhits [unlimited, so same as maxaccepts]
```

There's probably a sweet-spot between these defaults and the current exhaustive search, and I think `--maxaccepts 25 --maxrejects 100` might be pretty close.

thermokarst:
Thanks for posting those, @colinbrislawn, and thanks @nbokulich for sharing the "solid justification" with @colinbrislawn!

splaisan:
Unless I missed something, I do not see new user-settable parameters added for maxreject and maxhits announced end of lat year.
How can I get these in my running build please.

thermokarst:
Hi @splaisan, as you can see above, this is the pull request that proposes adding those parameters, and, it hasn't been merged yet. It is on our radar for the [2020.2 release](https://forum.qiime2.org/t/preview-qiime-2-2020-2-development-changelog/12328).

> How can I get these in my running build please.

You can install a [development version](https://dev.qiime2.org/latest/quickstart/) of QIIME 2, install this repo in development mode, and then pull down this PR locally. We cannot provide support on _how_ to do that, though, since it isn't a recommended user install method.

q2_feature_classifier/_vsearch.py
```diff
@@ -148,14 +165,33 @@ def classify_hybrid_vsearch_sklearn(ctx,
...
+                  'cases use maxhits to control the total number of hits '
+                  'returned. This option works in pair with maxrejects. '
+                  'The search process sorts target sequences by decreasing '
+                  'number of k-mers they have in common with the query '
+                  'sequence, using that information as a proxy for sequence '
+                  'similarity. After pairwise alignments, if the first target '
+                  'sequence passes the acceptation criteria, it is accepted '
+                  'as best hit and the search process stops for that query. '
+                  'If maxaccepts is set to a higher value, more hits are '
+                  'accepted. If --maxaccepts and --maxrejects are both set to '
```
```suggestion
                  'accepted. If maxaccepts and maxrejects are both set to '
```

I think leaving the `--` in there might be more confusing than beneficial.

nbokulich:
LGTM thanks! 🎾

PR #169, opened 2021-01-12
MAINT: Close #167

### LINKED ISSUE #167: --p-reads-per-batch should be made consistent with --p-confidence
**Improvement Description**
[This line](https://github.com/qiime2/q2-feature-classifier/blob/1edf5da73511225ac230053ee9524a42ece510d7/q2_feature_classifier/classifier.py#L236) should be changed to
```
    'reads_per_batch': Int % Range(0, None, inclusive_start=False) | Str % Choices(
            ['auto']),
```
to be explicit about when "auto" `reads_per_batch`ing is engaged. The appropriate changes should then be made downstream.

**Current Behavior**
```
  --p-reads-per-batch INTEGER
    Range(0, None)       Number of reads to process in each batch. If "auto",
                         this parameter is autoscaled to min( number of query
                         sequences / n-jobs, 20000).              [default: 0]
```
It isn't clear that the current default means "auto".

**Proposed Behavior**
Make it explicit that we're selecting auto reads per batch.

Oddant1:
Closes #167. I'm having some issues actually making a default value for the parameter work after changing the accepted values. It seems like regardless of how I change it I get the following
___
image
___

**Edit 1: I have tried a few things**

1. Changing the default value to 1, 'auto', or None. These all continued to produce the same error.
2. Changing the allowed types to `Range(1, None) | Str % Choices(['auto']`. Same error.
3. I tried keeping the allowed types the same as before and changing the default to -1 just to make sure that produced the same error as I have been getting. It did.

**Edit 2: I have identified the issue**
In this code right here we also default `reads_per_batch` to 0.
https://github.com/qiime2/q2-feature-classifier/blob/1edf5da73511225ac230053ee9524a42ece510d7/q2_feature_classifier/_vsearch.py#L68-L86
Later on down here when we actually register that method, we reuse the specification from `classify_sklearn`
https://github.com/qiime2/q2-feature-classifier/blob/1edf5da73511225ac230053ee9524a42ece510d7/q2_feature_classifier/_vsearch.py#L257-L266
So `classify_hybrid_vsearch_sklearn` was the method actually causing the issue not `classify_sklearn`. Should we change the default for that method as well or keep it as is?

**Aside**
Should we maybe make it so the error the framework raises (this one ```TypeError: Default value for parameter 'reads_per_batch' is not of semantic QIIME type Int % Range(1, None) | Str % Choices('auto') or `None`.```) says specifically which method is having a problem and maybe also what the invalid default value is? As it stands, the only hint from the error existed further up in the traceback, and that was only useful because the method that was actually causing problems was in a different file from the one I was messing with.


q2_feature_classifier/classifier.py
```diff
@@ -233,7 +233,8 @@ def classify_sklearn(reads: DNAFASTAFormat, classifier: Pipeline,
 
 
 _classify_parameters = {
-    'reads_per_batch': Int % Range(0, None),
+    'reads_per_batch': Int % Range(0, None, inclusive_start=False) | Str % Choices(
```
nbokulich, line 236: why not simplify to `Range(1, None) | Str % Choices(['auto'])`?

nbokulich (changes requested):
Thanks @Oddant1 !

I have one in-line comment on the change you made, but also a more general comment on the error you are seeing an other changes that are needed.

First, the error you are seeing is most likely because the default is still set to `0`, which is now an invalid value. Adjust that here:
https://github.com/qiime2/q2-feature-classifier/blob/66e4fd520c07cc1dea23a98e677fea1545127cc6/q2_feature_classifier/classifier.py#L202

Now that 0 is an invalid value, these lines also need to be adjusted:
https://github.com/qiime2/q2-feature-classifier/blob/66e4fd520c07cc1dea23a98e677fea1545127cc6/q2_feature_classifier/classifier.py#L208-L209

You should also check the unit tests to make sure `reads_per_batch` is not set to zero anywhere, I expect it might be.

q2_feature_classifier/classifier.py
Oddant1: I tried that, and I also tried changing the default from 0 to 1 or 'auto' or None. They all still produced the same error.

nbokulich (approved):
thanks @Oddant1 ! Changes look good — it looks like the unit tests are not triggering though, any clue why @Oddant1 or @thermokarst ?

thermokarst:
> it looks like the unit tests are not triggering though

Travis switched to a new credit-based system - we ran out of credits a long time ago. Ultimately we are going to move these all over to github actions, but not for another month or two. In the meantime I have been trying to coordinate purchasing more travis credits, but our admin team is pretty backed up. If you verify this manually/offline you should feel free to merge - busywork is still up and running.

nbokulich:
thanks @thermokarst ! if only travis issued a "build failed — give us $$$" notification.

testing locally now, will merge if success...

nbokulich:
@Oddant1 the tests are still failing, this is why:

>  So classify_hybrid_vsearch_sklearn was the method actually causing the issue not classify_sklearn. Should we change the default for that method as well or keep it as is?

yes, please change the default for that one to `auto` also. Thanks!

nbokulich (approved):
looks good, thanks @Oddant1 !

thermokarst:
@Oddant1  - do you think you can add a short entry for this change into the release changelog?

https://forum.qiime2.org/t/preview-qiime-2-2021-2-development-changelog/17815

PR #170, opened 2021-01-29
vsearch: options incompatible with --search_exact

emollier:
Good day,

When running the test suite on a system set up with _vsearch_ 2.15.1, the _test_consensus_assignment.py_ fails its test involving _search_exact_ with an error message including the following relevant part:

	Command: vsearch --search_exact /build/q2_feature_classifier/tests/data/se-dna-sequences.fasta --id 0.8 --query_cov 0.8 --strand both --maxaccepts 10 --maxrejects 0 --db /build/q2_feature_classifier/tests/data/se-dna-sequences.fasta --threads 1 --output_no_hits --blast6out /tmp/tmpomuor76d

	----------------------------- Captured stderr call -----------------------------
	Fatal error: Invalid options to command search_exact
	Invalid option(s): --id --maxaccepts --maxrejects --query_cov
	The valid options for the search_exact command are: --alnout --biomout --blast6out --bzip2_decompress --db --dbmask --dbmatched --dbnotmatched --fasta_width --fastapairs --gzip_decompress --hardmask --log --match --matched --maxhits --maxqsize --maxqt --maxseqlength --maxsizeratio --maxsl --mincols --minqt --minseqlength --minsizeratio --minsl --mintsize --mismatch --mothur_shared_out --no_progress --notmatched --notrunclabels --otutabout --output_no_hits --qmask --quiet --relabel --relabel_keep --relabel_md5 --relabel_self --relabel_sha1 --rowlen --samheader --samout --self --sizein --sizeout --strand --threads --top_hits_only --uc --uc_allhits --userfields --userout --xee --xsize

This commit alters the _vsearch_ command issued by _classify_consensus_vsearch_ by removing the options, and their argument, incompatible with _--search_exact_.

---

The issue has been caught initially when integrating the package into Debian and discussed alongside Bug#974839[1].

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=974839#24

Please note that I'm not competent with _vsearch_, so I most likely overlooked other options to set for an optimum usage, but you probably want to at least be aware of the issue.

Have a nice day,  :)

### DIFF
--- a/q2_feature_classifier/_vsearch.py
+++ b/q2_feature_classifier/_vsearch.py
@@ -50,6 +50,12 @@ def classify_consensus_vsearch(query: DNAFASTAFormat,
            '--threads', str(threads)]
     if search_exact:
         cmd[1] = '--search_exact'
+        # Remove options, and their matching argument, which are not compatible
+        # a --search_exact in recent versions of vsearch (such as 2.15.1).
+        for pop_arg in ['--id', '--maxaccepts', '--maxrejects', '--query_cov']:
+            pop_idx = cmd.index(pop_arg)
+            cmd.pop(pop_idx)
+            cmd.pop(pop_idx)
     if top_hits_only:
         cmd.append('--top_hits_only')
     if output_no_hits:


nbokulich:
Thanks @emollier !

Right now vsearch is pinned to <=2.7.0, though, so these changes are not currently relevant (I think those options must have been dropped in a more recent release of vsearch)
https://github.com/qiime2/q2-feature-classifier/blob/1edf5da73511225ac230053ee9524a42ece510d7/ci/recipe/meta.yaml#L27-L28

Looks like that comment line is outdated at this point as more recent builds are available in conda, but the issue is a bit more complicated than that...See also relevant discussion here:
https://github.com/qiime2/q2-vsearch/issues/75#issuecomment-658215857



updating to support more recent vsearch releases sounds like a good idea, but then I think we need to:
1. determine a minimum vsearch version to pin to if your changes break with older releases
2. unpin and fix in other QIIME 2 plugins (q2-vsearch, others?)
3. instead of silently dropping the now-invalid options when `search_exact=True`, this should be handled in the parameter registration to raise an error if invalid options are explicitly set by the user, by utilizing a `TypeMap` to specify valid combinations. This would be ideal so that users are aware of what is/is not happening under the hood. Then in the function instead of removing those parameters when `search_exact=True`, it would be better to only add those parameters when `search_exact=False`

emollier:
Hi Nicholas,

Thanks for the background information and your thoughts.  I see if I can come up with something more suitable along the lines of your point 3; I agree the present implementation is not ideal.

Have a nice day,  :)
Étienne.

nbokulich:
Hi @emollier this PR is over a year old now, and is still blocked for the foreseeable future, so I will close this. Thanks!

PR #176, opened 2022-03-22
refactored consensus classifiers

### LINKED ISSUE #152: `classify-consensus-*`: output alignment results report?
**Improvement Description**
Optionally output alignment results report from blast and vsearch classifiers as a metadata QZV. This report can be useful for QC purposes.

**Current Behavior**
The reports are generated, used to identify consensus taxonomy, but then thrown away. The user cannot access the intermediate file.

**Proposed Behavior**
Output as a metadata QZV the intermediate file that is parsed here: https://github.com/qiime2/q2-feature-classifier/blob/master/q2_feature_classifier/_consensus_assignment.py#L29

**Questions**
1. Any way to make this an optional output?
2. I suppose this implies converting these methods to pipelines (rhetorical question) 🤔

nbokulich:
Introduces BREAKING CHANGES

closes #152 

Quite a bit going on here:
1. big general refactor for improved readability and efficiency
2. Separated blast + vsearch aligners from the consensus classification step. Now we have thin wrappers for database searching, as well as a separate action for generating consensus annotations from BLAST6Format search results. Yep, that's 3 new actions for free!
3. `classify-consensus-*`  methods are now pipelines, linking together the separate actions mentioned above.
4. The Search results are output as part of the `classify-consensus-*` pipelines for convenient post-mortem. Re: #152 
5. New + improved unit tests etc (still more should be done on that front, but save it for a future PR).

This leaves a few major questions:
1. should the new `blast` action be generalized to support `blastn`, `blastp`, and maybe `blastx` in a single action? Or should we create separate actions for all 3?
2. Would it be better to output the blast-formatted alignment search results (current setup), a visualization (`metadata tabulate`) for convenient inspection, or both? re: #152 

I must admit that I have not tested extensively on real data yet, so maybe someone could kick the tires a bit on this one to see how it does in real life?


nbokulich:
small update: ran some runtime comparisons on the new vs. old blast consensus classifiers — negligible runtime difference since the majority of compute time is spent by `blastn` and the consensus step is relatively minor. So no CO2 emissions savings with this PR! But easier to read and maintain, and also better modularity.

lina-kim (commented):
Thanks @nbokulich! Looks great, and ran quickly too. A few comments:
* The `FeatureData[BLAST6]` file could not be visualized with `metadata tabulate` (as suggested in comments) - I ended up using `tools extract` instead
* I'd like search results as a QZV. Did you keep them as artifacts for downstream use?

nbokulich:
Thanks @lina-kim ! 
> The FeatureData[BLAST6] file could not be visualized with metadata tabulate

Oh, looks like we are missing a transformer! I have opened an issue for that, and can fix soon (unless if you would like to try tackling it): https://github.com/qiime2/q2-types/issues/273

> I'd like search results as a QZV. Did you keep them as artifacts for downstream use?

Yes the ouptut from the new `blast` and `vsearch` wrappers is currently a `FeatureData[BLAST6]` exactly for this purpose, and the transformer is the last missing piece (currently you can view as a `pd.DataFrame` to inspect the contents). The `classify_consensus_*` actions only output the taxonomy assignments right now, though, not the intermediate search results. Would you like me to add that as an output so that search results can be inspected alongside taxonomy classifications? (now that I type this out it sounds like a pretty good idea to me too, but just want to make sure I understand before committing new changes)

lina-kim:
> > The FeatureData[BLAST6] file could not be visualized with metadata tabulate
> 
> Oh, looks like we are missing a transformer! I have opened an issue for that, and can fix soon (unless if you would like to try tackling it): [qiime2/q2-types#273](https://github.com/qiime2/q2-types/issues/273)

Oh interesting, I'd be happy to give that a shot! Please feel free to assign that issue to me.

> > I'd like search results as a QZV. Did you keep them as artifacts for downstream use?
> 
> Yes the ouptut from the new `blast` and `vsearch` wrappers is currently a `FeatureData[BLAST6]` exactly for this purpose, and the transformer is the last missing piece (currently you can view as a `pd.DataFrame` to inspect the contents). The `classify_consensus_*` actions only output the taxonomy assignments right now, though, not the intermediate search results. Would you like me to add that as an output so that search results can be inspected alongside taxonomy classifications? (now that I type this out it sounds like a pretty good idea to me too, but just want to make sure I understand before committing new changes)

I hadn't considered all of the intermediate results, but the more metrics the merrier! By "search results" I meant the BLAST6-formatted assignments, but am definitely happy to have more outputs.

lina-kim (approved):

nbokulich:
hey @thermokarst will you be able to review this this week? Pls re-assign if needed.

thermokarst:
I'll take a look at it first thing this morning!

thermokarst (changes requested):
A few tests are failing when run against a newer vsearch. I can dig into it deeper if you need a hand, but time is tight this week.

thermokarst:
I'm guessing that something has changed with the output format, the obs results for the `test_vsearch_global` look like this

```
    qseqid   sseqid  pident  length  ...  sstart   send  evalue  bitscore
0  1111561  1111561   100.0    75.0  ...     1.0  150.0    -1.0       0.0
1  1111561  4424871    94.7    75.0  ...     1.0  150.0    -1.0       0.0
2  1111561  4476738    93.3    75.0  ...     1.0  150.0    -1.0       0.0
3   835097   835097   100.0    80.0  ...     1.0  150.0    -1.0       0.0
4     junk  4295019   100.0    88.0  ...     1.0  150.0    -1.0       0.0
5     junk  4295019   100.0    88.0  ...     1.0  150.0    -1.0       0.0
6     junk  4388133    94.2    86.0  ...     1.0  150.0    -1.0       0.0
7     junk  4477561    92.7    82.0  ...     1.0  150.0    -1.0       0.0
8     junk  4434148    90.1    81.0  ...     1.0  150.0    -1.0       0.0
```

instead of this

```
    qseqid   sseqid  pident  length  ...  sstart   send  evalue  bitscore
0  1111561  1111561   100.0    75.0  ...     1.0  150.0    -1.0       0.0
1   835097   835097   100.0    80.0  ...     1.0  150.0    -1.0       0.0
2     junk        *     0.0     0.0  ...     0.0    0.0    -1.0       0.0
```

I'm not familiar enough with this format to know what is or isn't acceptable behavior.

nbokulich:
thanks @thermokarst ! I will dig in.

nbokulich:
Okay looks like the change in output is due to updates in `vsearch v2.13.0` back in 2019, to improve handling of ambiguous characters. So technically the new outputs are "correct" (i.e., this is the expected behavior of the newer versions vsearch).

However, I have mixed feelings on this. The new hits that are detected are all basically to really bad reference seqs that are full of ambiguous characters. So you can note that the `junk` query sequence (which is a long string of `A`s) has fairly high pid and coverage scores to several reference sequences... which all happen to be garbage ref seqs (long tails of `N`s). So this is technically correct if `N` really means "any nucleotide" but in refseq dbs the `N` more often means "some nucleotide". So this boils down to more of a database quality issue (which could be addressed in other ways) and `q2-feature-classifier` might not be the hill for fighting this battle. But given the common occurrence of seqs with many ambiguous characters in common marker gene ref datasets, this is a change that _will_ impact analysis results, possibly in quite significant ways.

I would like to hear some opinions on this, maybe also from @mikerobeson and @misialq as well as @thermokarst ... the main options I see:
1. we resist change to preserve _the old ways_ (there is not an easy way it seems, e.g., no `maxN` parameter)
2. we just accept the altered functionality but document this in the release notes (i.e., that users should expect changes to the functionality and should carefully curate their reference  databases, e.g., with RESCRIPt). For the purposes of this PR, this would just require updating the exp results for these two failing tests and moving on.

I prefer option 2 but just wanted to share my dyspepsia over ambiguous bases.

Note that this would not be the only place that this functionality is altered. `q2-vsearch` and other plugins using `vsearch --search_global` (maybe also other functions) will also be affected. Again, this change is technically correct... so this should maybe be noted loudly in a blanket statement in the release notes for the next Q2 release, that various plugins/actions are impacted by the vsearch unpin and different results might be observed (e.g., also with OTU clustering).

mikerobeson:
I think we should go with option 2. This is related to a draft tutorial for the https://github.com/bokulich-lab/RESCRIPt/pull/140 PR. Which I recently shared with @nbokulich. In that tutorial, I warn that users need to be diligent about remove sequences with too many ambiguous bases, otherwise they'll increase the chances of extracting spurious sequence data due to these technically correct matches.

ebolyen:
It looks like this is the most relevant issue on vsearch: https://github.com/torognes/vsearch/issues/393

For 2, users would just need to run `rescript cull-seqs` right? That doesn't seem like a huge burden. If needed we could create some semantic properties to force the matter, but it may not be worth it. Especially if @mikerobeson's tutorial outlines what to do.

nbokulich:
Thanks @mikerobeson and @ebolyen for your feedback! Right, it is a simple fix with `cull-seqs` but users would need to be diligent (and read the docs). If that issue is ever fixed in vsearch we can just expose a relevant option here (and in other plugins) to toggle how Ns are handled. So "option 2" it is...

The failing tests are meant to also test cases where there are no hits (i.e., to the `junk` query) so in the end instead of changing the expected output I have trimmed the reference sequences to reduce the length of ambiguous regions to more realistic amounts (because otherwise this is impossible when there are garbage reference seqs consisting almost entirely of `N`s). To compensate, I have added a separate test to allow more permissive hits (so that `junk` still hits garbage with shorter ambiguous stretches). This will give us a test case to use later on if we ever expose an option to toggle `MaxN`.

thermokarst (approved):
LGTM! I'll leave merging up to @ebolyen.

nbokulich:
Hey @ebolyen is this ready to merge or do you need more time? This is blocking at least one other PR at the moment. Thanks!

nbokulich:
hey @ebolyen can this be merged? it would be great to get this in this month's release.

ebolyen:
Thanks @nbokulich, these changes will be in the August release! :tada:

PR #181, opened 2022-06-22
maint: revised vsearch signatures

thermokarst:
now that vsearch is unpinned, we need to revise how we call out to the tool. please see here:

https://github.com/qiime2/package-integration/runs/6996623476?check_suite_focus=true

for details on what exactly failed.


nbokulich:
LGTM thanks @thermokarst !

I think the parameter description(s) need to be updated to reflect this change, but I can open a separate PR now to work on this.

PR #189, opened 2023-04-25
Reduce the memory footprint of `find_consensus_annotation`

### LINKED ISSUE #188: Higher RAM use in classify-consensus-vsearch version 2022.8+
**Bug Description**
A [forum user reports](https://forum.qiime2.org/t/problem-with-feature-classifier-classify-consensus-vsearch/25368/6?u=nicholas_bokulich) higher RAM use with `classify-consensus-vsearch` in version 2022.8, but not in 2022.2

I have not confirmed this yet, but opening this ticket to remind me or anyone interested to investigate.

2022.8 introduced some significant changes in #176 that were intended to increase efficiency, but by decoupling the alignment and consensus classification steps it is possible that [pure speculation here] RAM usage is increasing because the alignment results are also written out as an intermediate output.

**References**
[forum x-ref](https://forum.qiime2.org/t/problem-with-feature-classifier-classify-consensus-vsearch/25368/6?u=nicholas_bokulich)

vaamb:
Sometimes my analyses crash while computing consensus annotation. I found out that it was due to an oom error, and that the culprit was pandas `replace` in `_compute_consensus_annotations` that could eat a significant amount of RAM (more than the 12 GB that I allocated on my WSL, for a dataset that was far less than 1 Go). To circumvent this issue, I used `iterrows` that reduces the memory footprint at a marginal cost in speed

This commit does not add new unit tests

### DIFF
--- a/q2_feature_classifier/_consensus_assignment.py
+++ b/q2_feature_classifier/_consensus_assignment.py
@@ -1,5 +1,5 @@
 # ----------------------------------------------------------------------------
-# Copyright (c) 2016-2022, QIIME 2 development team.
+# Copyright (c) 2016-2023, QIIME 2 development team.
 #
 # Distributed under the terms of the Modified BSD License.
 #
--- a/q2_feature_classifier/_consensus_assignment.py
+++ b/q2_feature_classifier/_consensus_assignment.py
@@ -85,8 +85,10 @@ plugin.methods.register_function(
 
 
 def _blast6format_df_to_series_of_lists(
-        assignments, ref_taxa,
-        unassignable_label=DEFAULTUNASSIGNABLELABEL):
+        assignments: pd.DataFrame,
+        ref_taxa: pd.Series,
+        unassignable_label: str = DEFAULTUNASSIGNABLELABEL
+) -> pd.Series:
     """import observed assignments in blast6 format to series of lists.
 
     assignments: pd.DataFrame
--- a/q2_feature_classifier/_consensus_assignment.py
+++ b/q2_feature_classifier/_consensus_assignment.py
@@ -105,7 +105,8 @@ def _blast6format_df_to_series_of_lists(
     # validate that assignments are present in reference taxonomy
     # (i.e., that the correct reference taxonomy was used).
     # Note that we drop unassigned labels from this set.
-    missing_ids = set(assignments['sseqid'].values) - set(ref_taxa.index) - {'*', ''}
+    missing_ids = \
+        set(assignments['sseqid'].values) - set(ref_taxa.index) - {'*', ''}
     if len(missing_ids) > 0:
         raise KeyError('Reference taxonomy and search results do not match. '
                        'The following identifiers were reported in the search '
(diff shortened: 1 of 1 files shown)


q2_feature_classifier/_consensus_assignment.py
```diff
@@ -116,8 +116,13 @@ def _blast6format_df_to_series_of_lists(
     # accession ID, so we will add this mapping to the reference taxonomy.
     ref_taxa['*'] = unassignable_label
     # map accession IDs to taxonomy
-    taxa_hits.replace(ref_taxa, inplace=True)
+    for index, value in assignments.iterrows():
+        sseqid = assignments.iloc[index]['sseqid']
+        assignments.at[index, 'sseqid'] = ref_taxa.at[sseqid]
```
ebolyen, line 121: This will ultimately cause `search_results` in `find_consensus_annotation` to be mutated.

That's probably fine, but something to keep track of. 
Would it be too much to do a df.copy() first? 
(I realize that memory pressure is the entire purpose of this change, so maybe it would be too much.)

ebolyen: Otherwise, as best I can tell. This does the same thing as before. So once our tests run, we should see the same results.

ebolyen (commented):
Thank you so much @vaamb! 

@nbokulich, do you have time to verify that these changes work as expected?

nbokulich:
Thanks so much for this @vaamb ! and thanks @ebolyen for the ping — looking at this has been on my to-do list.

@vaamb could you please address the lint errors and @ebolyen 's review first so that my review/test is on the updated version?

q2_feature_classifier/_consensus_assignment.py
```diff
@@ -99,13 +101,11 @@ def _blast6format_df_to_series_of_lists(
...
         the "assignment" input.
-    '''
-    taxa_hits = assignments.set_index('qseqid')['sseqid']
-
+    """
     # validate that assignments are present in reference taxonomy
     # (i.e., that the correct reference taxonomy was used).
     # Note that we drop unassigned labels from this set.
-    missing_ids = set(taxa_hits.values) - set(ref_taxa.index) - {'*', ''}
+    missing_ids = set(assignments['sseqid'].values) - set(ref_taxa.index) - {'*', ''}
```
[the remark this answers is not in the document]
vaamb: My bad, I'm used to Black that tends to be a bit more liberal than the pep 8

`pandas_obj.unique()` might be used and should not change the results but won't it need to still be wrapped in `set()` in order to work ? Making the line even bigger (although possibly faster)

q2_feature_classifier/_consensus_assignment.py
vaamb: Good idea! It should be possible to use a `df.copy()` copy at a minimal cost. The problem with `replace` is that somehow it was using exponential memory (a reported issue for this pandas version)

q2_feature_classifier/_consensus_assignment.py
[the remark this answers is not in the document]
ebolyen: No worries! And the `unique()` suggestion wasn't so much the line length as it might be a touch faster. You are right that the set subtraction would need to be handled differently in that case (or end up being basically the same as not doing it).

That's not important for this PR though :)

lizgehret:
Hey @vaamb just FYI I'm going to bump this PR to our 2023.8 project board since our PR cutoff date for 2023.5 is May 5th.

gregcaporaso:
@nbokulich - just wanted to ping you on this for a review when you get a chance.

nbokulich (approved):
sorry for the long delay in review. This looks good! I have tested and this results in a ~20% reduction in memory load and small decrease in runtime (N=1 test with GTDB as the reference).  Thanks for making these changes @vaamb !

vaamb:
@nbokulich no problem. Happy this little change is useful !

PR #195, opened 2023-12-13
Improve confidence-based taxonomy prediction performance

SirDavidLudwig:
This PR significantly improves the efficiency and performance of confidence-based taxonomy prediction by removing the need for `deepcopy`, resulting in a ~750% speed-up to sequences/second measured on our machine.

### DIFF
--- a/q2_feature_classifier/_skl.py
+++ b/q2_feature_classifier/_skl.py
@@ -7,7 +7,6 @@
 # ----------------------------------------------------------------------------
 
 from itertools import islice, repeat
-from copy import deepcopy
 
 from joblib import Parallel, delayed
 
--- a/q2_feature_classifier/_skl.py
+++ b/q2_feature_classifier/_skl.py
@@ -20,7 +20,7 @@ class _TaxonNode:
     # labels. It allows one to quickly find class label indices of taxonomy
     # labels that satisfy a given taxonomy hierarchy. For example, given the
     # 'k__Bacteria' taxon, the _TaxonNode.range property will yield all class
-    # label indices where 'd__Bacteria' is a prefix.
+    # label indices where 'k__Bacteria' is a prefix.
 
     name: str
     offset_index: int
--- a/q2_feature_classifier/_skl.py
+++ b/q2_feature_classifier/_skl.py
@@ -75,13 +75,13 @@ def _predict_chunk_with_conf(pipeline, separator, confidence, chunk):
     split_classes = [c.split(separator) for c in pipeline.classes_]
     for seq_id, taxon, class_probs in zip(seq_ids, y, prob_pos):
         taxon = taxon.split(separator)
-        classes = zip(deepcopy(split_classes), class_probs)
+        classes = zip(split_classes, class_probs)
         result = []
-        for level in taxon:
-            classes = [cls for cls in classes if cls[0].pop(0) == level]
-            cum_prob = sum(c[1] for c in classes)
+        for rank, level in enumerate(taxon):
+            cum_prob = sum(cls[1] for cls in classes if cls[0][rank] == level)
             if cum_prob < confidence:
                 break
+            classes = zip(split_classes, class_probs)
             result.append(level)
             result_confidence = cum_prob
         if result:
--- a/q2_feature_classifier/_skl.py
+++ b/q2_feature_classifier/_skl.py
@@ -13,40 +13,44 @@ from typing import Dict, List
 
 from joblib import Parallel, delayed
 
-
+# The _TaxonNode is used to build a hierarchy from a list of sorted class
+# labels. It allows one to quickly find class label indices of taxonomy labels
+# that satisfy a given taxonomy hierarchy. For example, given the
+# 'k__Bacteria' taxon, the _TaxonNode.range property will yield all class
+# label indices where 'd__Bacteria' is a prefix.
 @dataclass
-class _TaxonPredictionNode:
+class _TaxonNode:
     name: str
-    offset: int # The offset index of this taxon in the probability vector
-    children: Dict[str, "_TaxonPredictionNode"] = field(default_factory=dict, repr=False)
+    offset_index: int # The offset index of this taxon in the class list
+    children: Dict[str, "_TaxonNode"] = field(
+        default_factory=dict,
+        repr=False)
 
     @classmethod
-    def create_tree(cls, classes: List[str]):
+    def create_tree(cls, classes: List[str], separator: str):
+        if not all(a <= b for a, b in zip(classes, classes[1:])):
+            raise Exception("classes must be in sorted order")
         root = cls("Unassigned", 0)
-        for i, label in enumerate(classes):
-            taxons = label.split(';')
+        for class_start_index, label in enumerate(classes):
+            taxons = label.split(separator)
             node = root
             for name in taxons:
                 if name not in node.children:
-                    node.children[name] = cls(name, i)
+                    node.children[name] = cls(name, class_start_index)
                 node = node.children[name]
         return root
 
     @property
     def range(self) -> range:
-        return range(self.offset, self.offset + len(self))
+        return range(
+            self.offset_index,
+            self.offset_index + self.num_leaf_nodes)
 
     @cached_property
     def num_leaf_nodes(self) -> int:
         if len(self.children) == 0:
             return 1
         return sum(c.num_leaf_nodes for c in self.children.values())
-
-    def __len__(self) -> int:
-        return self.num_leaf_nodes
-
-    def __getitem__(self, key: str) -> "_TaxonPredictionNode":
-        return self.children[key]
 
 
 _specific_fitters = [
@@ -109,7 +113,7 @@ def _predict_chunk_with_conf(pipeline, separator, confidence, chunk):
 
     y = pipeline.classes_[prob_pos.argmax(axis=1)]
 
-    taxonomy_tree = _TaxonPredictionNode.create_tree(pipeline.classes_)
+    taxonomy_tree = _TaxonNode.create_tree(pipeline.classes_, separator)
 
     results = []
     for seq_id, taxon, class_probs in zip(seq_ids, y, prob_pos):
@@ -119,7 +123,7 @@ def _predict_chunk_with_conf(pipeline, separator, confidence, chunk):
         result = []
         current = taxonomy_tree
         for rank in split_taxon:
-            current = current[rank]
+            current = current.children[rank]
             cum_prob = class_probs[current.range].sum()
             if cum_prob < confidence:
                 break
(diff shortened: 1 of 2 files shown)


q2_feature_classifier/_skl.py
```diff
@@ -75,13 +74,13 @@ def _predict_chunk_with_conf(pipeline, separator, confidence, chunk):
     split_classes = [c.split(separator) for c in pipeline.classes_]
     for seq_id, taxon, class_probs in zip(seq_ids, y, prob_pos):
         taxon = taxon.split(separator)
-        classes = zip(deepcopy(split_classes), class_probs)
+        classes = zip(split_classes, class_probs)
```
BenKaehler: Suggest changing this to
`classes = list(zip(split_classes, class_probs))`
and deleting line 83.

q2_feature_classifier/_skl.py
```diff
@@ -75,13 +74,13 @@ def _predict_chunk_with_conf(pipeline, separator, confidence, chunk):
...
         result = []
-        for level in taxon:
-            classes = [cls for cls in classes if cls[0].pop(0) == level]
-            cum_prob = sum(c[1] for c in classes)
+        for rank, level in enumerate(taxon):
+            cum_prob = sum(cls[1] for cls in classes if cls[0][rank] == level)
```
BenKaehler: This is the potential error. The error arises because sometimes taxa match at a later lever (in the label) but differ earlier. For instance the Ruminococcus genus exists in both the Lachnospiraceae and Ruminococcaceae families in Greengenes.

I suggest matching on all the levels up to the current one. Something like
```
for rank in range(1, len(taxon) + 1):
    cum_prob = sum(cls[1] for cls in classes if cls[0][:rank] == taxon[:rank])
```

q2_feature_classifier/_skl.py
```diff
@@ -75,13 +74,13 @@ def _predict_chunk_with_conf(pipeline, separator, confidence, chunk):
...
             if cum_prob < confidence:
                 break
+            classes = zip(split_classes, class_probs)
```
BenKaehler: Suggest deleting this line and replacing the original definition of classes with a list rather than a zip object.

I'm fairly confident that the result would be neater, less confusing, and more efficient.

I might be wrong about the efficiency, though, and it might be faster to create the zip object at every iteration of this loop than to create the whole list before iterating. You have done performance tests so you might like to test both.

q2_feature_classifier/_skl.py
```diff
@@ -75,13 +74,13 @@ def _predict_chunk_with_conf(pipeline, separator, confidence, chunk):
...
             result.append(level)
```
BenKaehler: Suggest replacing this line with
`result = taxon[:rank]`
if you make the suggested change at line 80.

BenKaehler (changes requested):
Hi @SirDavidLudwig, thanks very much for this PR. That improvement in efficiency is great.

There is one potential error (although it doesn't show up in the unit tests) and I have one suggestion. Please refer to the comments that I attached to the code.

SirDavidLudwig:
If we can safely make the assumption that the model's classes are ordered alphabetically (which appears to be the case, but I could be wrong), I'd like to propose a new tree based solution to this problem. Not only does it address the issue you mentioned, but the search space is heavily optimized, allowing for a 100X speed-up. I attached a simple bar plot comparing some numbers that I observed in my testing. If this assumption is safe to make, I'd like to update this PR to reflect these new changes.

image

SirDavidLudwig:
I didn't realize Github automatically updates pull requests when code is committed. I can submit this idea as a different pull request and revert back to the original changes we discussed.

q2_feature_classifier/_skl.py
SirDavidLudwig: Hmm... This makes sense, but the redundant checking will also have a significant performance impact. We could go with this for now though, but I believe there's a better solution that could be implemented.

q2_feature_classifier/_skl.py
SirDavidLudwig: The change in efficiency would be negligible here... But I agree, it would be cleaner to create a list instance once in the beginning of the loop.

q2_feature_classifier/_skl.py
```diff
@@ -6,11 +6,49 @@
...
+from typing import Dict, List
 
 from joblib import Parallel, delayed
 
+
+@dataclass
+class _TaxonPredictionNode:
+    name: str
+    offset: int # The offset index of this taxon in the probability vector
+    children: Dict[str, "_TaxonPredictionNode"] = field(default_factory=dict, repr=False)
```
BenKaehler: I think the linter is failing somewhere around here. Please fix.

q2_feature_classifier/_skl.py
```diff
@@ -6,11 +6,49 @@
...
+
+    @classmethod
+    def create_tree(cls, classes: List[str]):
+        root = cls("Unassigned", 0)
+        for i, label in enumerate(classes):
+            taxons = label.split(';')
```
BenKaehler: Please pass the `separator` that is passed to `predict` down to this level, so that it is guaranteed to be consistent.

q2_feature_classifier/_skl.py
```diff
@@ -71,25 +109,26 @@ def _predict_chunk_with_conf(pipeline, separator, confidence, chunk):
 
     y = pipeline.classes_[prob_pos.argmax(axis=1)]
 
+    taxonomy_tree = _TaxonPredictionNode.create_tree(pipeline.classes_)
```
BenKaehler: As you state in the comments, you assume that `pipeline.classes_` is sorted. The sklearn classes I've looked at do sort their classes, but please add an assertion here to check that it is not violated in the future.

BenKaehler (changes requested):
Thanks @SirDavidLudwig, using the natural order of the classes in the probability vector is a clever idea. That should also fix the issue I raised with the last version of this PR.

Please address the minor comments that I've raised and I will approve.

BenKaehler:
Hi @SirDavidLudwig, just an afterthought: could you please also add a brief "here be dragons"-style comment to the new `_TaxonPredicitonNode` class to explain to future maintainers why it works?

q2_feature_classifier/_skl.py
SirDavidLudwig: Since the tree structure requires the classes to be sorted, I put the assertion inside the `create_tree` method.

q2_feature_classifier/_skl.py
```diff
@@ -6,11 +6,55 @@
...
 from joblib import Parallel, delayed
 
+
+@dataclass
+class _TaxonNode:
+    # The _TaxonNode is used to build a hierarchy from a list of sorted class
+    # labels. It allows one to quickly find class label indices of taxonomy
+    # labels that satisfy a given taxonomy hierarchy. For example, given the
+    # 'k__Bacteria' taxon, the _TaxonNode.range property will yield all class
+    # label indices where 'd__Bacteria' is a prefix.
```
BenKaehler: Sorry @SirDavidLudwig, did you mean k__Bacteria here?

BenKaehler (changes requested):
Thanks @SirDavidLudwig, your changes look great. Just one minor typo in your comment that I think needs fixing. Thanks again.

q2_feature_classifier/_skl.py
SirDavidLudwig: Yes, sorry about that. Just pushed the fix.

BenKaehler:
Hi @ebolyen, are you ok with me merging this, or should it get another review? Thanks!

misialq (changes requested):
Hey @SirDavidLudwig, thanks for your contribution! It would be super lovely if you could ad a couple of unit tests specific to your new data class, demonstrating the usage of the class methods/properties and the edge cases they are trying to account for. Thanks!

BenKaehler:
Sorry for the delay @SirDavidLudwig!

After being unable to contact Evan, I decided to do a couple more tests.

I ran the Silva 138.1 515f-806r-animal-distal-gut-classifier off Zenodo on 2,689 sequences from the PD Mouse dataset with 11 cores on my aging laptop.

It took 1m47s using the classifier proposed here and 6m14s using the existing classifier, so that's a substantial improvement, even with the unfairness of parallel computing and the latency of loading a large classifier 11 times.

All taxonomic classifications were identical and the largest difference in confidence estimates was 9.6e-14.

Thanks very much @misialq for running another set of eyes over the changes. You are, of course, correct that more unit tests should be added. I think once those are in we're definitely ready to merge.

These are my suggestions for some unit tests that could be added to the end of `ClassifierTests` in `test_classifier.py` (thanks copilot):

``` python
    def test_TaxonNode_create_tree(self):
        classes = ['a;b;c', 'a;b;d', 'a;e;f', 'a;e;g']
        separator = ';'
        tree = _TaxonNode.create_tree(classes, separator)
        self.assertEqual(
            tree.children['a'].children['b'].children['c'].name, 'c')
        self.assertEqual(
            tree.children['a'].children['b'].children['d'].name, 'd')
        self.assertEqual(
            tree.children['a'].children['e'].children['f'].name, 'f')
        self.assertEqual(
            tree.children['a'].children['e'].children['g'].name, 'g')
    
    def test_TaxonNode_range(self):
        classes = ['a;b;c', 'a;b;d', 'a;e;f', 'a;e;g']
        separator = ';'
        tree = _TaxonNode.create_tree(classes, separator)
        self.assertEqual(
            tree.children['a'].children['b'].children['c'].range, range(0, 1))
        self.assertEqual(
            tree.children['a'].children['b'].children['d'].range, range(1, 2))
        self.assertEqual(
            tree.children['a'].children['e'].children['f'].range, range(2, 3))
        self.assertEqual(
            tree.children['a'].children['e'].children['g'].range, range(3, 4))
        self.assertEqual(
            tree.children['a'].children['b'].range, range(0, 2))
        self.assertEqual(
            tree.children['a'].children['e'].range, range(2, 4))
    
    def test_TaxonNode_num_leaf_nodes(self):
        classes = ['a;b;c', 'a;b;d', 'a;e;f', 'a;e;g']
        separator = ';'
        tree = _TaxonNode.create_tree(classes, separator)
        self.assertEqual(tree.num_leaf_nodes, 4)
        self.assertEqual(tree.children['a'].num_leaf_nodes, 4)
        self.assertEqual(tree.children['a'].children['b'].num_leaf_nodes, 2)
        self.assertEqual(tree.children['a'].children['e'].num_leaf_nodes, 2)
```

SirDavidLudwig:
I added the above unit tests to `ClassifierTests` as suggested. Let me know if there's anything else. Thanks!

BenKaehler (approved):
Perfect. Thanks @SirDavidLudwig!

PR #199, opened 2024-03-25
BUG: update help classify-sklearn action and help text for unsupporte…

colinvwood:
Fixes a mistake made in the previous release where I had assumed that negative values would make it to this action. In fact they're intercepted by the framework now. We can't support the "n less than total" logic here anymore so this is an interface change.


gregcaporaso (approved):
Looks good! @colinvwood, just be sure to add a note to the release notes about the interface change (mention that it was a bug, which is why there wasn't advance notice for the interface change).

PR #210, opened 2025-01-31
BUG: nondeterministic inclusion of sequences in extract-reads

ebolyen:
Discovered by: https://forum.qiime2.org/t/qiime-feature-classifier-extract-reads-successfully-extracts-different-number-of-reads-each-time/32437

The issue seems to be that degenerate bases are a `set`, which means they get a different order between Python sessions:
https://github.com/scikit-bio/scikit-bio/blob/main/skbio/sequence/_dna.py#L188-L198

As a consequence, when a terrible alignment exists, this code:
```python
if best_score is None or score > best_score:
            best_score = score
```
goes with the first alignment found (which is nondeterministic). Since it's a bad alignment, none of the other concrete sequences in the degenerate list can beat it. So, depending on the specific identity of that first alignment, the sequence will be retained or rejected.

Essentially, only terrible alignments will be occasionally included, based on an (un)lucky order of the degenerate map.

We should do something other than this PR to solve the issue, since I don't think we want these bad alignments anyhow. But worst case, this will at least make the bad alignments appear the same way across different runs.

Another option, which seems to work is increasing `identity` which causes the bad alignments to be filtered out (even if they got lucky before).

### DIFF
--- a/q2_feature_classifier/_cutter.py
+++ b/q2_feature_classifier/_cutter.py
@@ -25,7 +25,7 @@ def _seq_to_regex(seq):
     for base in str(seq):
         if base in skbio.DNA.degenerate_chars:
             result.append('[{0}]'.format(
-                ''.join(skbio.DNA.degenerate_map[base])))
+                ''.join(sorted(skbio.DNA.degenerate_map[base]))))
         else:
             result.append(base)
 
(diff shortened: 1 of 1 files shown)


lizgehret:
test failures caused by rna 2.2.0. fix in progress [here](https://github.com/qiime2/distributions/pull/589).

gregcaporaso:
I'm reviewing this and thinking we should go with this fix for now. We discussed (IRL) that we need to (primarily) use score here, not identity, as score integrates the length of the match while identity doesn't, but that identity could be a secondary filter to remove low quality matches. scikit-bio is planning to move away from SSW, so we're going to need to update this code at that time anyway, so I propose that we just deal with the non-deterministic issue for now, and come up with a new solution that doesn't use SSW. At that point, a benchmark of the approach and methods for "cutting" is in order. Perhaps the right approach at that time integrates a minimum match percentage and a minimum percent identity across the full length of the match. 

cc @nbokulich

gregcaporaso:
I pushed an empty commit here to trigger the tests.

ebolyen:
I think that's a good plan. Let's revisit this with new aligners.

PR #212, opened 2025-02-07
PIN: template out vsearch version

lizgehret:
paired PR in distributions: https://github.com/qiime2/distributions/pull/600


colinvwood:
waiting on qiime2/q2-feature-table#329 to be available

PR #217, opened 2025-08-02
LANG: 2025.10 update & pin templating

lizgehret:


gregcaporaso:
Note that the unit tests are passing for my updates to the aligner, but I'm still working on some real-world testing of these changes. I'll add some notes here as I go. As we discussed, we're still a little way off from merging these PRs across the distros - I should be done by the time that's all ready to go. 

Update: I did some experimenting and documented the updates - all of that is [here](https://gist.github.com/gregcaporaso/1f00f93f967331a764a6940e5b464472) for now. I think the changes to `_cutter.py` are good to go, but I'm going to share with a few folks and will circle back if that changes. 

Another update: I did some experimenting with the `identity` parameter today, and I do think we should drop it to 0.70 as I think with the new code we're not extracting some reads that do have reasonable alignments. I'm seeing some reads that are no longer extracted though that I can't currently explain (they forward and reverse alignments result in sequences that exceed my identity threshold), so we should add some additional unit tests of the new functionality here before these changes are merged to make sure that it's all working as expected. A possible reason for this is how the output of my new function interacts with the trimming/truncating/length parameters (e.g., am I dropping the primers from the extracted reads too soon).

gregcaporaso:
Closing in favor of #220

PR #220, opened 2026-03-18
LANG: 2026.4 update

gregcaporaso:
This replaces #217. The following is copied from that discussion:

> Note that the unit tests are passing for my updates to the aligner, but I'm still working on some real-world testing of these changes. I'll add some notes here as I go. As we discussed, we're still a little way off from merging these PRs across the distros - I should be done by the time that's all ready to go.
> 
> Update: I did some experimenting and documented the updates - all of that is [here](https://gist.github.com/gregcaporaso/1f00f93f967331a764a6940e5b464472) for now. I think the changes to _cutter.py are good to go, but I'm going to share with a few folks and will circle back if that changes.
>
> Another update: I did some experimenting with the identity parameter today, and I do think we should drop it to 0.70 as I think with the new code we're not extracting some reads that do have reasonable alignments. I'm seeing some reads that are no longer extracted though that I can't currently explain (they forward and reverse alignments result in sequences that exceed my identity threshold), so we should add some additional unit tests of the new functionality here before these changes are merged to make sure that it's all working as expected. A possible reason for this is how the output of my new function interacts with the trimming/truncating/length parameters (e.g., am I dropping the primers from the extracted reads too soon).

Still to-do:
- [x] update default value for identity parameter
- [x] add new unit tests as described above

### DIFF
--- a/q2_feature_classifier/_cutter.py
+++ b/q2_feature_classifier/_cutter.py
@@ -103,8 +103,6 @@ def _match_percent(primer, target):
         elif (primer_c in skbio.DNA.degenerate_chars and
               target_c in skbio.DNA.degenerate_map[primer_c]):
             matches += 1
-        else:
-            continue
     return matches / len(primer)
 
 
--- a/q2_feature_classifier/_cutter.py
+++ b/q2_feature_classifier/_cutter.py
@@ -8,6 +8,7 @@
 
 import skbio
 import os
+import numpy as np
 from joblib import Parallel, delayed, effective_n_jobs
 
 from qiime2.plugin import Int, Str, Float, Range, Choices
@@ -37,54 +38,6 @@ def _primers_to_regex(f_primer, r_primer):
                                _seq_to_regex(r_primer.reverse_complement()))
 
 
-def _local_aln(primer, sequence):
-    best_score = None
-    for one_primer in sorted([str(s) for s in primer.expand_degenerates()]):
-        # `sequence` may contain degenerates. These will usually be N
-        # characters, which SSW will score as zero. Although undocumented, SSW
-        # will treat other degenerate characters as a mismatch. We acknowledge
-        # that this approach is a heuristic to finding an optimal alignment and
-        # may be revisited in the future if there's an aligner that explicitly
-        # handles degenerates.
-        this_aln = \
-            skbio.alignment.local_pairwise_align_ssw(skbio.DNA(one_primer),
-                                                     sequence)
-        score = this_aln[1]
-        if best_score is None or score > best_score:
-            best_score = score
-            best_aln = this_aln
-    return best_aln
-
-
-def _semisemiglobal(primer, sequence, reverse=False):
-    if reverse:
-        primer = primer.reverse_complement()
-
-    # locally align the primer
-    (aln_prim, aln_seq), score, (prim_pos, seq_pos) = \
-        _local_aln(primer, sequence)
-    amplicon_pos = seq_pos[1]+len(primer)-prim_pos[1]
-
-    # naively extend the alignment to be semi-global
-    bits = [primer[:prim_pos[0]], aln_prim, primer[prim_pos[1]+1:]]
-    aln_prim = ''.join(map(str, bits))
-    bits = ['-'*(prim_pos[0]-seq_pos[0]),
-            sequence[max(seq_pos[0]-prim_pos[0], 0):seq_pos[0]],
-            aln_seq,
-            sequence[seq_pos[1]+1:amplicon_pos],
-            '-'*(amplicon_pos-len(sequence))]
-    aln_seq = ''.join(map(str, bits))
-
-    # count the matches
-    matches = sum(s in skbio.DNA.degenerate_map.get(p, {p})
-                  for p, s in zip(aln_prim, aln_seq))
-
-    if reverse:
-        amplicon_pos = max(seq_pos[0]-prim_pos[0], 0)
-
-    return amplicon_pos, matches, len(aln_prim)
-
-
 def _exact_match(seq, f_primer, r_primer):
     try:
         regex = _primers_to_regex(f_primer, r_primer)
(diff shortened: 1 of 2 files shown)


q2_feature_classifier/_cutter.py
```diff
@@ -95,12 +48,100 @@ def _exact_match(seq, f_primer, r_primer):
...
+                    else:
+                        sm[(row, col)] = mismatch
+                else:  # degenerate char in query sequence is always a mismatch
+                    sm[(row, col)] = mismatch
+            else:  # primer character is degenerate
+                if c2 in skbio.DNA.degenerate_map[c1]:
+                    sm[(row, col)] = match
+                else:
+                    sm[(row, col)] = mismatch
+    return skbio.SubstitutionMatrix(chars, sm)
```
colinvwood, line 92: The fact that rows are primers and columns queries is left implicit. Even if we don't encode that in the `SubstitutionMatrix` it could be helpful to mention in the docstring.

q2_feature_classifier/_cutter.py
```diff
@@ -95,12 +48,100 @@ def _exact_match(seq, f_primer, r_primer):
...
+
+
+def _match_percent(primer, target):
```
colinvwood, line 95: what are the types of `primer`, `target`?

q2_feature_classifier/_cutter.py
```diff
@@ -95,12 +48,100 @@ def _exact_match(seq, f_primer, r_primer):
...
+    for primer_c, target_c in zip(str(primer), str(target)):
+        if target_c in skbio.DNA.degenerate_chars:
+            continue
+        if primer_c == target_c:
+            matches += 1
+        elif (primer_c in skbio.DNA.degenerate_chars and
+              target_c in skbio.DNA.degenerate_map[primer_c]):
+            matches += 1
+        else:
+            continue
```
colinvwood:
```suggestion
```

q2_feature_classifier/_cutter.py
```diff
@@ -95,12 +48,100 @@ def _exact_match(seq, f_primer, r_primer):
...
+            else:  # primer character is degenerate
+                if c2 in skbio.DNA.degenerate_map[c1]:
+                    sm[(row, col)] = match
+                else:
+                    sm[(row, col)] = mismatch
+    return skbio.SubstitutionMatrix(chars, sm)
+
+
+def _match_percent(primer, target):
+    """ Compute percent of matching positions in alignments, accounting for
```
colinvwood:
```suggestion
    """ Compute proportion of matching positions in alignments, accounting for
```

q2_feature_classifier/_cutter.py
```diff
@@ -95,12 +48,100 @@ def _exact_match(seq, f_primer, r_primer):
...
+    return matches / len(primer)
+
+
+def _align_primer(primer, target, substitution_matrix, reverse=False):
+    if reverse:
+        primer = primer.reverse_complement()
+
+    # perform pairwise semi-global alignment, such that gaps on the
+    # ends of primer aren't scored but gaps on the ends of target are
+    # scored.
```
colinvwood: Is this correct? Seems like we want to penalize if the primer falls off the end of the target, but not if the primer begins past the beginning of (within) the target.

colinvwood: The values passed to `free_ends` below seem to confirm what I'm thinking

q2_feature_classifier/_cutter.py
```diff
@@ -95,12 +48,100 @@ def _exact_match(seq, f_primer, r_primer):
...
+    # but degenerate characters in target are always considered
+    # mismatches
+    aln = skbio.alignment.pair_align_nucl(
+        primer, target, mode='global', sub_score=substitution_matrix,
+        free_ends=[True, True, False, False], trim_ends=True)
+    msa = skbio.TabularMSA.from_path_seqs(aln.paths[0], (primer, target))
+    match_percent = _match_percent(msa[0], msa[1])
+
+    if reverse:
+        amplicon_pos = aln.paths[0].starts[1]
```
colinvwood, line 149: Is amplicon position exclusive of the primer?

q2_feature_classifier/_cutter.py
```diff
@@ -95,12 +48,100 @@ def _exact_match(seq, f_primer, r_primer):
...
-    if (b_matches + e_matches) / (b_length + e_length) >= identity:
-        return seq[beg:end]
-    return None
+    substitution_matrix = _create_asymmetric_primer_substitution_matrix()
+    amp_start, f_match_percent = \
+        _align_primer(f_primer, seq, substitution_matrix)
+    amp_end, r_match_percent = \
+        _align_primer(r_primer, seq, substitution_matrix, reverse=True)
+    if f_match_percent >= identity and r_match_percent >= identity:
+        return seq[amp_start:amp_end]
```
colinvwood, line 163: Is `amp_end` already exclusive? Seems like we would need a `+ 1` otherwise. Guessing the tests would have caught this though.

colinvwood (commented):
This is only for `_cutter.py` so far, I'll read the tests next

q2_feature_classifier/_cutter.py
gregcaporaso: they're both assumed to be skbio.DNA objects. I added this in the doc string.

q2_feature_classifier/tests/test_cutter.py
```diff
@@ -151,3 +158,194 @@ def test_extract_reads_fail_min_len_greater_than_trunc_len(self):
...
+        # rc('TTTT') = AAAA, which matches the AAAA region perfectly
+        primer = skbio.DNA('TTTT')
+        target = skbio.DNA('GGGGAAAAGGGG')
+        _, match_pct = _align_primer(primer, target, self.sm, reverse=True)
+        self.assertAlmostEqual(match_pct, 1.0)
+
+    def test_partial_forward_match_percent(self):
+        # AAAC aligns to AAAA in target — 3 of 4 positions match
+        primer = skbio.DNA('AAAC')
+        target = skbio.DNA('GGGGAAAAGGGG')
```
colinvwood, line 278: In cases like this do we care whether the returned alignment is 

AAAC
AAAA

or

AAAC
AAAG

when they score equally? Like should we prefer shorter/longer amplicons?

q2_feature_classifier/_cutter.py
gregcaporaso: i think i have this right, but my language might not be clear. does this example clarify? if so, i'll throw it in there in place of what i have: 

```
gaps on the ends of the primer, as in the following, are free from penalization:
--AAAA----------			 
CCAAAAGGGGCCCCTT
or
----------CCCC--			 
CCAAAAGGGGCCCCTT

gaps on the end of the target, as in the following, incur gap penalty:
AAAA------
--AAGGGGCC

or 
------CCCC
AAGGGGCC--
```

q2_feature_classifier/_cutter.py
gregcaporaso: yep, confirmed

q2_feature_classifier/_cutter.py
gregcaporaso: yep, confirmed

q2_feature_classifier/tests/test_cutter.py
colinvwood: Just curious, not suggesting we add any changes to prefer one over the other

q2_feature_classifier/_cutter.py
colinvwood: Yes, this clarifies. I just misinterpreted the original comment

q2_feature_classifier/tests/test_cutter.py
gregcaporaso: Hmm, given that this is used for training classifiers, I think we would want to prefer the longer amplicon.

colinvwood (approved):

colinvwood:
@gregcaporaso I'm finished with changes & review, everything (all of feature-classifier) passed in the docker container for me

gregcaporaso:
Sounds good, thanks so much @colinvwood. I'll restart the tests for @lizgehret, and @lizgehret this should be good to merge whenever you're ready!

PR #221, opened 2026-04-25
NEW: add report to `extract-reads` enabling interrogation of why reads were/were not extracted

gregcaporaso:
This report will be helpful for determining whether some changes I want to make to this action impact the results, and additionally can help to explore primer bias and other features expected amplicons (e.g., pass a GTDB 16S reference database in along with a primer pair, and then `metadata tabulate` the report with the reference taxonomy to explore whether taxa of interest are likely to be amplified, and/or how long the resulting amplcions should be, etc).


gregcaporaso:
Replaced by #222.

